PR #1, opened 2022-03-11
Add ability to play videos from library

itsWindows11:


Screenbox/App.xaml.cs
```diff
@@ -123,7 +123,7 @@ protected override void OnLaunched(LaunchActivatedEventArgs e)
             if (rootFrame.Content == null)
             {
                 SetMinWindowSize();
-                rootFrame.Navigate(typeof(PlayerPage));
+                rootFrame.Navigate(typeof(MainPage));
```
huynhsontung, line 126: Due to the current LibVLC limitation, there can only be one instance of the media player element per `LibVLC` instance. This means the player will break if you navigate away from the `PlayerPage`.

You can see there is an inline Frame in `PlayerPage` that loads the `MainPage` to circumvent this.

itsWindows11: > Due to the current LibVLC limitation, there can only be one instance of the media player element per `LibVLC` instance. This means the player will break if you navigate away from the `PlayerPage`.
> 
> You can see there is an inline Frame in `PlayerPage` that loads the `MainPage` to circumvent this.

I think you can init the player in the app initialization code and then that would fix it. You could revert that if you want for now.

huynhsontung:
Rebased and merged in 7a61e371d7a8b9b850e46428229c82442abddf80

PR #56, opened 2023-04-07
feat: File Associations improvements

United600:
Made some improvements to file associations, Logos and DisplayNames with localization in mind.

Feel free do discard the icons if you don't like them, tried my best to keep the color palette, they're kinda hard to see in dark mode or maybe not.

Also I think you should split the files DisplayName to another resource file to be better organized and easier to expand.

Cheers.


Screenbox/Package.appxmanifest
```diff
@@ -50,45 +50,234 @@
       </uap:VisualElements>
       <Extensions>
         <uap:Extension Category="windows.fileTypeAssociation">
-          <uap:FileTypeAssociation Name="videos">
+          <uap:FileTypeAssociation Name="ThirdGeneration">
```
huynhsontung: These names are invalid. No uppercase characters are allowed. Consider using snake case or kebab case names.

huynhsontung (changes requested):
I really like these icons. I look forward to getting this merged! The friendly names can stay in the `Resources.resw` for now. I can deal with that later.

United600 (commented):
Didn't knew about it, If you don't have a preference I'll use kebab case. The documentation wasn't clear to me, so I only avoided numbers.

I have a feeling that I'm also missing something.

Screenbox/Assets/FileAssociationFLAC.targetsize-16.png
huynhsontung, line 1: At this size, the upper part (the flag) seems a bit too big and heavy. Maybe we can make the flag's corners less rounded.

Screenbox/Assets/FileAssociationFLAC.targetsize-48.png
huynhsontung, line 1: Sizes 48 and 32 have the same flag issue.

Screenbox/Assets/FileAssociationFLAC.targetsize-512.png
huynhsontung, line 1: There are some shadow issues at the left edge. I have attached the logo background so you don't have to redo the background again.

[logo.zip](https://github.com/huynhsontung/Screenbox/files/11183534/logo.zip)

huynhsontung (changes requested):
Nice work! There are some minor issues that I noticed when testing this. Personally, it feels a bit odd to have a custom icon shape for videos but not for audio. But I don't have any other idea right now and it still looks nice.

Can you also share your design file so new file types can be added in the future?

Screenbox/Strings/en-US/Resources.resw
```diff
@@ -534,4 +534,70 @@
...
+  </data>
+  <data name="WebMediaFileFriendlyName" xml:space="preserve">
+    <value>WebM Video</value>
+  </data>
+  <data name="WindowsMediaAudioFileFriendlyName" xml:space="preserve">
+    <value>Windows Media Audio</value>
+  </data>
+  <data name="WindowsMediaVideoFileFriendlyName" xml:space="preserve">
+    <value>Windows Media Video</value>
+  </data>
```
huynhsontung: I'm not sure if these names are widely recognizable. For example, "Audio Video Interleave" doesn't strike me as AVI at glance. By default, Windows uses `<extension> File` as the display name (e.g. "FLAC File", "AVI File"). If we override this, we will have to commit to localizing all display names in the future. Not sure if it's worth the effort, especially when many of these names are foreign to most people.

United600: > Nice work! There are some minor issues that I noticed when testing this. Personally, it feels a bit odd to have a custom icon shape for videos but not for audio. But I don't have any other idea right now and it still looks nice.

The audio it is custom (cd/vinyl album), I used the logo shapes just so it wasn't a big black blob,

>Can you also share your design file so new file types can be added in the future?

Sure I'll share the svg here. Sorry, had to remove the shadows because I was having problems.
FileAssociationAudio
FileAssociationVideo

Screenbox/Assets/FileAssociationFLAC.targetsize-16.png
United600: It's supposed to be a musical note, I can reduce the size.

Screenbox/Assets/FileAssociationFLAC.targetsize-512.png
United600: I'll use it. Thanks.

Screenbox/Strings/en-US/Resources.resw
United600: > I'm not sure if these names are widely recognizable. For example, "Audio Video Interleave" doesn't strike me as AVI at glance. By default, Windows uses `<extension> File` as the display name (e.g. "FLAC File", "AVI File"). If we override this, we will have to commit to localizing all display names in the future. Not sure if it's worth the effort, especially when many of these names are foreign to most people.

Some of them can be shortened, like the AVI and FLAC, but before (in the Windows Media Player times) used "proper" extension files. AVI was 'Video Clip', 3GPP was '3GPP Audio/Video' for example, now in the new Media Player (and Groove before) they do basically nothing and Windows just uses `<extension> File`.

Screenbox/Assets/FileAssociationFLAC.targetsize-16.png
huynhsontung: image

I meant the flag part looks a bit too thick and too rounded. The musical note size is fine.

huynhsontung (approved):
Looks good 👍

United600:
Oops just checked your mock-up audio icons and they look slightly better, if I have time I'll try to replicate the look

PR #67, opened 2023-05-05
Improvements to globalization

United600:
Moved manifest strings to another resource file for better organization, should help with localization.

Also added some strings (eg. "AppShortName", "AppDisplayName") for other writing systems.

#### ToDo
- [ ] Move file logos assets to another folder


huynhsontung:
There are reports that file association icons are causing slowdown for some users. I will take care of them and move them to another folder.

United600:
> I think this change is good. Moving these strings to a different resource file would remove codegen for them as well. The strings are only used in the manifest and not in code so I think it's safe. Thank you.

Yes, easier to handle.

> There are reports that file association icons are causing slowdown for some users. I will take care of them and move them to another folder.

Pretty sure it's not related to that but tbh not entirely happy with the icons, at least on 150 scale the 32px video downscales really bad. On this weekend I'll see what I can do.
Maybe don't even need one for each extension, with thumbnails I hardly see the bigger ones but it's always useful on list view, food for thought.

huynhsontung:
> The 32 pixel downscaled at 150 scale.

I suspect that icon is being dynamically downscaled. Since 32 at 100% is still quite legible. See #68 for more icon sizes that I have generated. Anything below 32px is going to be the simple icon.

United600:
yeah. I did a quick check for List View, so it goes something like this:

| Scale | Size |
|-------|------|
| 100 | 16 |
| 125 | 20 |
| 150 | 24 |
| 175 | 26 or 26? |
| 200 | 32 |

huynhsontung:
> | 175 | 26 or 28? |

Native applications don't use 26 and 28. How did you verify this?

United600:
> > | 175 | 26 or 28? |
> 
> Native applications don't use 26 and 28. How did you verify this?

just changed the scale, log off, log in, open explorer.

and yes nobody uses 26 or 28, all downscaled the 32 pixel, my guess 175 is rarely used

huynhsontung:
That's fine. Thank you for verifying. We will stick with what native applications use. They are: 

128, 129, 16, 20, 24, 256, 30, 32, 336, 36, 40, 48, 60, 63, 64, 72, 80, 96

See #68 for more details.

With 5 main variations. The rest are scaled versions of these 5.

Screenbox/Package.appxmanifest
```diff
@@ -218,7 +218,7 @@
         </uap:Extension>
         <uap:Extension Category="windows.fileTypeAssociation">
           <uap:FileTypeAssociation Name="quicktime">
-            <uap:DisplayName>ms-resource:QuickTimeFileFriendlyName</uap:DisplayName>
+            <uap:DisplayName>mms-resource:ManifestResources/FileDisplayNameQuickTime</uap:DisplayName>
```
huynhsontung: I think you made a typo here. `mms-resource`

scripts/update-manifest.ps1
```diff
@@ -12,7 +12,7 @@ $ManifestPath = "$ProjectPath\Package.appxmanifest"
 $xmlDoc.Package.Identity.Name="18496Starpine.Screenbox"
 $xmlDoc.Package.Identity.Publisher="CN=ABCDF790-DBE4-48F7-8204-32FCB69ADF9C"
 $xmlDoc.Package.Properties.DisplayName="Screenbox"
-$xmlDoc.Package.Applications.Application.VisualElements.DisplayName="Screenbox"
+$xmlDoc.Package.Applications.Application.VisualElements.DisplayName="ms-resource:ManifestResources/AppDisplayName"
```
huynhsontung, line 15: Does AppDisplayName need to be localized?

huynhsontung (commented):
Looks good. We can merge once you resolve the typo.

Screenbox/Package.appxmanifest
United600: Fixed

scripts/update-manifest.ps1
United600: not right now, just future proofing for other alphabets

PR #68, opened 2023-05-06
Optimize file icons

huynhsontung:
Provide more file icon sizes to avoid dynamic sizing, which leads to poor performance in Explorer.


United600:
don't you think `"Assets\Icons\Audio"` path is a little too much?

what about simply `"Assets\FileIcons"` and then reduce the filename to just `{ext}.png`

United600:
also I think we can improve the 24px icon, what you think about these? I've changed the video design a little bit, more rounded corners throughout the icon.

Audio targetsize-24
Video targetsize-24

huynhsontung:
> don't you think `"Assets\Icons\Audio"` path is a little too much?
> 
> what about simply `"Assets\FileIcons"` and then reduce the filename to just `{ext}.png`

I want to group different icon designs into separate folders so that we know all icons in one folder have the same design. I don't mind any naming convention, as long as it is consistent.

> also I think we can improve the 24px icon, what you think about these? I've changed the video design a little bit, more rounded corners throughout the icon.

We can revisit any redesign in future releases and another PR. For this one, I'm just addressing the performance issue in explorer.

PR #107, opened 2023-07-11
Add Chinese translation.

okyan123:
Hi, following the translation method you mentioned, I have successfully added Chinese support to this software. I have tested it on my computer and also had friends help with testing to ensure there are no issues.

However, I'm not very familiar with UWP software development, so currently, it seems like the software's language is changing according to the user's computer system settings. Perhaps I should add an option to change the language in the software settings, but I'm not sure how to do that 😂.

This is a screenshot of the effect comparison 
7a650a35e839eb5e5e31e9f4ed8feff5

### DIFF
--- a/Screenbox.sln
+++ b/Screenbox.sln
@@ -108,6 +108,7 @@ Global
 		HideSolutionNode = FALSE
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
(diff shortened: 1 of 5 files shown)


huynhsontung:
Thank you for your translation! I don't think we need a language button just yet, certainly something nice to have.

huynhsontung (approved):

PR #116, opened 2023-08-03
Improved library folder appearance in settings and added display name

itsWindows11:
Before:
image

After:
image


huynhsontung:
I'm not sure if this is a good idea. I think the path is the main identifier for the folder and not the name.

itsWindows11:
@huynhsontung Windows Media Player (the legacy one) shows library folders in this same way too in its library management dialog.

huynhsontung:
You are right, the legacy media player does use this layout. That's good enough for me. 👍

PR #145, opened 2023-08-26
chore: Consolidation of the WinUI xmlns namespace

United600:
Changed some WinUI namespaces to be consistent throughout project.


huynhsontung:
Thanks for the consolidation. LGTM. 👍

Issue #123, opened 2023-08-10
There is a major issue with the Chinese translation.😨


okyan123:
Hello again, it's me. Last month, I submitted the Chinese translation. A couple of days ago, I was thrilled to discover that you had released a new version with the Chinese translation included. However, upon launching the software, I encountered a rather significant issue. To put it simply, the Chinese translation was completely scrambled, or more precisely, the translations seemed to be displaced and appeared in incorrect places.

For example, here the {} is used in C# for string interpolation, hava a variable in between. However, this sentence is now in the wrong position, causing {} to remain unchanged.
![image](https://github.com/huynhsontung/Screenbox/assets/100983934/04b69b85-0c3a-4dc0-9ccc-5189456978e6)

Even the software's name is incorrect. The term 'AAC Audio', which represents the software name, has been placed in the wrong location and has even replaced the actual software name.
![image](https://github.com/huynhsontung/Screenbox/assets/100983934/6ccaf1af-29bd-4997-9861-014f2b362a96)

I'm currently taking a look at the version history to find out how to address this issue. Do you have any thoughts?😭

okyan123:
I just tried downloading the release's sideload.zip and installing it. I encountered the same issue, and when I tried to open the software for the second time, it crashed. I couldn't open it successfully. It seems like there's a somewhat serious issue going on.😨

huynhsontung:
Oh no I will take a look ASAP. I have no idea what would cause this. The store-publish package is built in a [CI pipeline](https://github.com/huynhsontung/Screenbox/actions/runs/5791953947). Maybe there is some misconfiguration with the CI? I will try to build locally and publish a hotfix later tonight.

huynhsontung:
Can you try this side load package instead? I just built it locally.
https://github.com/huynhsontung/Screenbox/releases/download/v0.9.9/Screenbox_0.9.9.0_Test.zip

huynhsontung:
Can confirm that the package from the store has the issue. I couldn't replicate it with the sideload package though. Uninstalling the store version then install the sideload and it works as expected.

okyan123:
Thank you. With the sideload version, I succeeded😄, but another issue arose. Here's the situation: I have two computers. Yesterday, on my main computer, I encountered this issue. Both the store and sideload versions had this issue. Even after uninstalling the store version and installing the sideload version, the Chinese translation still went wrong.

[...]

PR #146, opened 2023-08-30
chore: Update to deprecated glyphs

United600:
According to @Jay-o-Way findings [here](https://github.com/microsoft/PowerToys/issues/10331), those deprecated glyphs ([Segoe Fluent Icons font](https://learn.microsoft.com/en-us/windows/apps/design/style/segoe-fluent-icons-font#icon-list)) cause issues with CJK languages.

Maybe affected #123.


Jay-o-Way:
You'll want to include these too:
image

United600:
> You'll want to include these too: image → `E768` and `E769`

good catch, thanks

huynhsontung (approved):
Confirmed everything works on Win10 and 11. Thanks for the update. 👍

PR #160, opened 2023-09-29
fix: add missing Composers property in PropertiesView

United600:
Adds the [composer property](https://learn.microsoft.com/en-us/uwp/api/windows.storage.fileproperties.musicproperties.composers?view=winrt-22621) for audio files to the PropertiesView dialog. I've provisionally reused the writers string.

### Example
Properties view dialog with the writers property highlighted

### Task
- [x] Create a Composers string


huynhsontung:
Added "Composers" string

image

PR #166, opened 2023-10-02
fix: audio subtitle picker header style

huynhsontung:
image

### DIFF
--- a/Screenbox/Controls/AudioTrackSubtitlePicker.xaml
+++ b/Screenbox/Controls/AudioTrackSubtitlePicker.xaml
@@ -23,7 +23,8 @@
                     <TextBlock
                         Margin="15,0,0,0"
                         VerticalAlignment="Center"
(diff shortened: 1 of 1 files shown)


Screenbox/Controls/AudioTrackSubtitlePicker.xaml
```diff
@@ -23,7 +23,8 @@
                     <TextBlock
                         Margin="15,0,0,0"
```
United600, line 24: Can you try "16,0,0,0"?

huynhsontung: Here is 16,0,0,0. I think it aligns better.

image

huynhsontung: Updated margin to "16,0,0,0" and pushed to main.

PR #177, opened 2023-10-07
feat: drag anywhere on the player page to move window

### LINKED ISSUE #127: Drag the window anywhere instead of just the title bar

huynhsontung:
Fixes #127


huynhsontung:
Turns out this causes more problems than originally expected. Pointer events are not captured under the element set as the title bar. This breaks player context menu and controls auto hide logic.

PR #180, opened 2023-10-12
fix: better content navigation and minor a11y improvements

United600:
Some improvements to navigation and xaml consistency, here some of the changes:

- xml namespaces
- left side gutter is no longer a deadspace
- some buttons were missing accessibility tags
- converted some {x:Bind}
- better App.xaml readability

### DIFF
--- a/Screenbox/App.xaml
+++ b/Screenbox/App.xaml
@@ -8,21 +8,11 @@
     xmlns:toolkitConverters="using:Microsoft.Toolkit.Uwp.UI.Converters">
 
     <Application.Resources>
-        <ResourceDictionary>
-            <ResourceDictionary.MergedDictionaries>
-                <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls" />
-
+        <muxc:XamlControlsResources>
+            <muxc:XamlControlsResources.MergedDictionaries>
                 <ResourceDictionary>
-                    <!--  Override  -->
-                    <!--  NavigationView content background to enable Mica  -->
-                    <StaticResource x:Key="NavigationViewContentBackground" ResourceKey="SystemControlTransparentBrush" />
-                    <!--  NavigationView content border thickness  -->
-                    <Thickness x:Key="TopNavigationViewContentGridBorderThickness">0,0,0,0</Thickness>
-                    <!--  Top NavigationView side margins  -->
-                    <Thickness x:Key="TopNavigationViewTopNavGridMargin">0,0</Thickness>
-                    <!--  Top NavigationView height to match PageHeader line height  -->
-                    <x:Double x:Key="NavigationViewTopPaneHeight">44</x:Double>
 
+                    <!--  Media content background and foreground  -->
                     <ResourceDictionary.ThemeDictionaries>
                         <ResourceDictionary x:Key="Dark">
                             <SolidColorBrush x:Key="AccentListViewItemBackgroundBrush" Color="#08FFFFFF" />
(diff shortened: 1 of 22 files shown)


Screenbox/App.xaml
```diff
@@ -6,87 +6,129 @@
     xmlns:converters="using:Screenbox.Converters"
     xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
     xmlns:toolkitConverters="using:Microsoft.Toolkit.Uwp.UI.Converters">
+
     <Application.Resources>
-        <muxc:XamlControlsResources>
-            <muxc:XamlControlsResources.MergedDictionaries>
+        <ResourceDictionary>
+            <ResourceDictionary.MergedDictionaries>
+                <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls" />
```
huynhsontung: Is there any reason for this change? The [WinUI documentation](https://learn.microsoft.com/en-us/windows/apps/winui/winui2/getting-started) recommends the original usage.

Using the recommended method doesn't allow resource override from `App.xaml`. Are there any references for this behaviour?

Screenbox/App.xaml
```diff
@@ -6,87 +6,129 @@
...
+
+                <!--  Overwrite WinUI resources  -->
                 <ResourceDictionary>
+                    <!--  NavigationView content background to enable Mica  -->
+                    <StaticResource x:Key="NavigationViewContentBackground" ResourceKey="SystemControlTransparentBrush" />
```
huynhsontung: `NavigationViewContentBackground` follows the WinUI design language so I don't intend to change this value. We can introduce this when we have more appearance customization options.

Screenbox/Pages/MusicPage.xaml
```diff
@@ -67,33 +60,35 @@
...
+                    AutomationProperties.Name="{strings:Resources Key=AddFolder}"
                     Command="{x:Bind ViewModel.AddFolderCommand}"
                     Style="{StaticResource AccentButtonStyle}"
                     ToolTipService.ToolTip="{strings:Resources Key=AddMusicFolderToolTip}"
                     Visibility="{x:Bind helpers:SystemInformationExtensions.IsDesktop}">
-                    <StackPanel Orientation="Horizontal">
+                    <StackPanel Orientation="Horizontal" Spacing="8">
                         <FontIcon Glyph="&#xe8f4;" />
-                        <TextBlock Margin="8,0,0,0" Text="{x:Bind strings:Resources.AddFolder}" />
+                        <TextBlock Text="{strings:Resources Key=AddFolder}" />
```
huynhsontung: Sadly the `Spacing` property ignores element visibility. So, when the label is collapsed, the extra spacing is still visible.

image

Screenbox/Pages/HomePage.xaml
```diff
@@ -251,14 +248,15 @@
             <TextBlock
                 x:Name="GridViewHeaderText"
                 Grid.Row="0"
-                Margin="{StaticResource ContentPageRightMargin}"
+                Margin="{StaticResource PageContentMargin}"
+                AutomationProperties.HeadingLevel="Level2"
```
huynhsontung, line 258: Nice addition!

Screenbox/Pages/AllVideosPage.xaml
```diff
@@ -60,15 +61,15 @@
         <muxc:ProgressBar
             x:Name="LoadingIndicator"
             Margin="0,8,0,0"
-            Padding="{StaticResource ContentPageRightMargin}"
+            Padding="{StaticResource PageContentMargin}"
```
huynhsontung: Having both left and right padding is broken for `ProgressBar`.

https://github.com/huynhsontung/Screenbox/assets/31434093/62329210-f4c4-4d2b-a1a9-3e699529895a

Screenbox/Pages/MusicPage.xaml
```diff
@@ -104,77 +99,73 @@
...
+                        <TextBlock Text="{strings:Resources Key=ShuffleAndPlay}" />
                     </StackPanel>
                 </Button>
 
                 <muxc:ProgressBar
                     x:Name="LoadingIndicator"
                     Grid.Row="1"
                     Margin="0,16,0,0"
-                    Padding="{StaticResource ContentPageRightMargin}"
+                    Padding="{StaticResource PageContentMargin}"
```
huynhsontung: Same `ProgressBar` padding issue here.

Screenbox/Pages/MainPage.xaml
```diff
@@ -26,11 +26,8 @@
             <Thickness x:Key="AppTitleMarginMinimal">98,0,0,0</Thickness>
 
             <DataTemplate x:Key="SearchAutoSuggestBoxItemTemplate">
-                <StackPanel Orientation="Horizontal">
-                    <FontIcon
-                        Margin="0,0,8,0"
-                        FontSize="16"
```
huynhsontung, line 32: The font size is necessary to make the icon smaller. I see you already have a resource for `FontIcon` in `App.xaml` but it may not work if the icon is in a flyout or a dialog since they use a different visual root from the main application.

old:
image

new:
image

huynhsontung: Side note you can see the 16 font size is sharper than the default size on a 1920x1080 screen, 100% scaling.

huynhsontung (changes requested):
Thank you for these long overdue improvements. I spot a few minor issues.

Screenbox/App.xaml
United600: thanks for the link.
I took inspiration from the PowerToys repo, their [App.xaml](https://github.com/microsoft/PowerToys/blob/main/src/settings-ui/Settings.UI/SettingsXAML/App.xaml).

Screenbox/App.xaml
United600: Technically this is not a new change, the only "change" was the brush (it was "Transparent").
Moved these resources: [Music page](https://github.com/huynhsontung/Screenbox/blob/main/Screenbox/Pages/MusicPage.xaml#L108)/[Videos page](https://github.com/huynhsontung/Screenbox/blob/main/Screenbox/Pages/VideosPage.xaml#L117)

Screenbox/Pages/MusicPage.xaml
United600: How did I missed this... :-(

Screenbox/Pages/HomePage.xaml
United600: > Nice addition!

If you're talking about the AutomationProperties I sprinkled some fixes and improvements

Screenbox/Pages/AllVideosPage.xaml
United600: > Having both left and right padding is broken for `ProgressBar`.
> 
>  Recording.2023-10-11.220005.mp4

Another big miss, how do I fix this? Just tweak the margins like the breadcrumb bars?

Screenbox/Pages/MainPage.xaml
United600: > Side note you can see the 16 font size is sharper than the default size on a 1920x1080 screen, 100% scaling.

:-(

Screenbox/Pages/MusicPage.xaml
United600: I'll revert all of them, even the ones that don't make use of the visibility element

Screenbox/App.xaml
huynhsontung: I see. Please set the resource locally on Music and Videos pages instead since this change also makes the main NavigationView background transparent.

United600: Done. Also tweaked again the line height so now they both match.

huynhsontung (approved):
lgtm 👍

PR #189, opened 2023-10-16
feat: main NavigationView improvements

huynhsontung:
NavigationView now has a 48px margin from the top, following the WinUI guidelines.

image

Adjusted animation timing for nav bar expand and collapse.

https://github.com/huynhsontung/Screenbox/assets/31434093/3e25f749-778f-41c4-92f1-5200ccf43724


United600 (commented):
We probably shouldn't remove it, otherwise the Grid Headers and NavigationView headers will be different (eg. VideosPage - FolderViewPage) and it'll have cascading effects for alignment.
example

We can add height to the grid row as a fix, I guess.

Edit: Oops, the [commit](https://github.com/huynhsontung/Screenbox/pull/189/commits/244ec4a4376bfe181515cef51797e8aa132b471c#diff-2a65fda2e29f78abdd07a946580e19031e614196813d4484b8ba06beb1c58894).

United600 (commented):
Also maybe increase these values (I'm getting -148 for staying the same):
[AlbumArtBG](https://github.com/huynhsontung/Screenbox/pull/189/files#diff-08fca18ca93304b506c122bd2c9e960238d3719c25d281425349070daf58ef24L39)
[AlbumArtBG-Minimal](https://github.com/huynhsontung/Screenbox/pull/189/files#diff-08fca18ca93304b506c122bd2c9e960238d3719c25d281425349070daf58ef24L201)

[ArtistArtBG](https://github.com/huynhsontung/Screenbox/pull/189/files#diff-cb49ec8b9ba7f5eb3dfa816b3499739ea237528ca95d99db89c1895d1abd5969L112)
[ArtistArtBG-Minimal](https://github.com/huynhsontung/Screenbox/pull/189/files#diff-cb49ec8b9ba7f5eb3dfa816b3499739ea237528ca95d99db89c1895d1abd5969L271)

AlbumArt comparison

Well actually more art looks good 👌.

Screenbox/Controls/Styles/CustomNavigationView.xaml
```diff
@@ -0,0 +1,756 @@
...
+                            <Grid
+                                x:Name="ButtonHolderGrid"
+                                Grid.Row="1"
+                                Margin="{ThemeResource NavigationViewButtonHolderGridMargin}">
+                                <Button
+                                    x:Name="NavigationViewBackButton"
+                                    Width="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.SmallerPaneToggleButtonWidth}"
+                                    VerticalAlignment="Top"
+                                    IsEnabled="{TemplateBinding IsBackEnabled}"
+                                    Style="{StaticResource NavigationBackButtonNormalStyle}"
```
United600: On Windows 10 is using the "wrong style" (in this case the Windows 10 style).
Back button Windows 10

huynhsontung: Seems like I used the wrong NavigationView style. Will investigate.

huynhsontung:
Thanks for explaining the LineHeight usage. I wasn't sure why it was there before.

> We can add height to the grid row as a fix, I guess.

Perhaps MinHeight? We allow up to two lines for the page title so we would need the row height to be dynamic.

United600:
> Perhaps MinHeight? We allow up to two lines for the page title so we would need the row height to be dynamic.

I, personally, like LineHeight better. Easier to manage and simplifies our job (don't have to keep track of all headers). Other option is to override the NavigationView height, I did try that but 44 wasn't possible (LineHeight didn't reduce past 45 and something).

Screenbox/Controls/Styles/CustomNavigationView.xaml
United600: > Seems like I used the wrong NavigationView style. Will investigate.

Also confused, is it this [one](https://github.com/microsoft/microsoft-ui-xaml/blob/main/dev/NavigationView/NavigationBackButton_rs1_themeresources.xaml)?

United600:
> Perhaps MinHeight? We allow up to two lines for the page title so we would need the row height to be dynamic.

About that.... I think I removed most of them, only left MaxLines=2 for the search pages.

Screenbox/Controls/Styles/CustomNavigationView.xaml
huynhsontung: There were more complications than I initially thought with `CustonNavigationViewStyle` so I removed it and went back to modification in code after the template apply.

huynhsontung:
After some discussion, we agreed to use MinHeight to standardize header height everywhere.

PR #203, opened 2023-10-30
style: playback flyout ui

United600:
Combines and redesigns the various flyouts that can customize media playback settings.

**Mock-up**
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://github.com/huynhsontung/Screenbox/assets/698155/cbff01d2-0361-4576-b521-eea0924fcfbb">
  <source media="(prefers-color-scheme: light)" srcset="https://github.com/huynhsontung/Screenbox/assets/698155/845b977b-08c8-4c96-9461-fe0d15a26168">
  Mock-up of video playback options in light and dark mode.
</picture>

**V1 Tasks**
- [ ] Localize strings
- [ ] Aspect ratio ComboBox
- [ ] Disable sliders that don't apply to audio playback
- [ ] Override the minimum playback speed
- [ ] Add and Bind Reset command


United600:
> I would add the option to manually edit the value.

This was how I first envisioned, for the first version we can go without it.
PlaybackOptionsVideo Dark

But I think the plan is to end up with something like this.
PlaybackOptionsVideo Dark2

PR #236, opened 2023-11-18
style: change speed glyph

United600:
Allows, at a glance, to view how the playback speed is set on the context menu.
Switched to a more appropriate glyph.


huynhsontung:
I moved the glyph selection logic to the control code behind instead for simplicity.

PR #268, opened 2023-12-25
fix: incorrect color space when rendering HDR content on SDR screen

huynhsontung:
HDR contents appear blown out on an SDR screen. Fixed in LibVLC: https://github.com/huynhsontung/libvlc-nuget/pull/12

image


maisondasilva:
Just one question, why do these cuts appear? I don't see it in another play

image

huynhsontung:
@maisondasilva Those are chapter markers. They only show up when there is chapter information encoded in the media file. Kinda like YouTube Chapters features except you can't jump to one, yet.

PR #288, opened 2024-01-03
feat: improve gamepad interaction with listviews

United600:


Screenbox/App.xaml
```diff
@@ -117,7 +117,10 @@
 
                     <!--  Require focus engagement on a gamepad  -->
                     <!--  https://learn.microsoft.com/en-us/windows/apps/design/input/gamepad-and-remote-interactions#focus-engagement  -->
-                    <Style x:Key="ListViewItemFocusEngagementStyle" TargetType="ListViewItem">
+                    <Style
+                        x:Key="ListViewItemFocusEngagementStyle"
+                        BasedOn="{StaticResource DefaultListViewItemStyle}"
```
huynhsontung, line 122: This style needs to be based on the default ListViewItem style from WinUI to not break the default styling.

PR #294, opened 2024-01-07
feat(ui): add entrance animation for settings page

huynhsontung:


Screenbox/Pages/SettingsPage.xaml
```diff
@@ -113,6 +113,7 @@
                 Padding="{StaticResource PageContentMargin}">
                 <StackPanel.ChildrenTransitions>
                     <RepositionThemeTransition IsStaggeringEnabled="False" />
+                    <EntranceThemeTransition IsStaggeringEnabled="True" />
```
United600:
```suggestion
                    <EntranceThemeTransition FromVerticalOffset="50" IsStaggeringEnabled="True" />
```

Screenbox/Pages/SettingsPage.xaml
```diff
@@ -113,6 +113,7 @@
                 Padding="{StaticResource PageContentMargin}">
                 <StackPanel.ChildrenTransitions>
```
United600, line 114: Shouldn't both transitions be inside a ```<TransitionCollection>```?

huynhsontung: The `<TransitionCollection>` considers the whole panel, in this case `StackPanel`, as one entity to calculate the animation. So when the cards change their position, there won't be any animation since from the `RepositionThemeAnimation` perspective, the panel only changes in size and not position. 

`EntranceThemeAnimation` would work here, but `IsStaggeringEnabled` would not have the staggered animation effect.

huynhsontung: No need. It's implicitly defined. See [ChildrenTransitions ](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.panel.childrentransitions#remarks)

Screenbox/Pages/SettingsPage.xaml
huynhsontung: Is there any particular reason for this offset value?

United600: No particular reason, there seems to be a consensus on that value:
[1](https://github.com/CommunityToolkit/Windows/blob/main/components/SettingsControls/samples/SettingsPageExample.xaml#L32) [2](https://github.com/microsoft/calculator/blob/main/src/Calculator/Views/Settings.xaml#L80)

Screenbox/Pages/SettingsPage.xaml
United600: Good to know, thanks for the link

PR #330, opened 2024-02-28
chore: application resource dictionary cleanup

United600:
- Removed some duplicate converters
- Renamed SubtitleTextMargin 'x:Key' to something more global
- Removed PageHeaderMargin/Minimal and start using a combination of "PageContentMargin/Minimal" and "BottomMediumMargin" to achieve the same result


Screenbox/App.xaml
```diff
@@ -18,15 +18,12 @@
...
                         <ResourceDictionary x:Key="Default">
-                            <StaticResource x:Key="ListViewAltItemBackgroundBrush" ResourceKey="CardBackgroundFillColorDefaultBrush" />
                             <StaticResource x:Key="AccentFillColorCustomBrush" ResourceKey="SystemAccentColorLight2" />
                         </ResourceDictionary>
                         <ResourceDictionary x:Key="Light">
-                            <StaticResource x:Key="ListViewAltItemBackgroundBrush" ResourceKey="CardBackgroundFillColorDefaultBrush" />
                             <StaticResource x:Key="AccentFillColorCustomBrush" ResourceKey="SystemAccentColorDark1" />
                         </ResourceDictionary>
                         <ResourceDictionary x:Key="HighContrast">
-                            <StaticResource x:Key="ListViewAltItemBackgroundBrush" ResourceKey="SystemColorButtonFaceColorBrush" />
```
huynhsontung, line 29: Doesn't removing the last resource break HighContrast?

United600: It shouldn't, the CardBackgroundFillColorDefaultBrush is the "same". IIRC it's ButtonFace (currently) vs WindowColor.

huynhsontung: Confirmed

PR #351, opened 2024-04-01
style: stretch indeterminate progress bars to full width of the page

United600:


Screenbox/App.xaml
```diff
@@ -88,6 +88,8 @@
                     <Thickness x:Key="PageThumbnailMargin">52,0</Thickness>
                     <Thickness x:Key="PageThumbnailMarginMinimal">14,0</Thickness>
 
+                    <Thickness x:Key="TopMediumMargin">0,12,0,0</Thickness>
+                    <Thickness x:Key="TopLargeMargin">0,16,0,0</Thickness>
```
huynhsontung, line 92: I'm a bit confused. When do we use Medium or Large margin? Is there a pattern for when to use each one?

United600: Not really and we should. I'll have to backtrack to see why but it seems to me that I started with those two values but then after testing touch targets the Large key fitted better. Maybe the Medium key stuck around because of the BreadcrumbBar, at least on the VideosPage.

Screenbox/Pages/FolderListViewPage.xaml
```diff
@@ -78,8 +78,6 @@
 
     <Grid>
         <muxc:ProgressBar
-            x:Name="LoadingProgressBar"
-            Margin="{StaticResource PageContentMargin}"
```
United600, line 82: @huynhsontung Can you check this spacing with the ListView? I pretty sure the ListView (like the GridView on the FolderViewPage) needs a Margin property, unfortunately I can't test it.

Screenbox/App.xaml
United600: Can confirm that the VideosPage medium margin is there to match [this](https://github.com/huynhsontung/Screenbox/blob/main/Screenbox/Pages/FolderViewPage.xaml#L73).

The other use case is more subjective, the "center" between the 'Shuffle & Play' button and the music content.

Screenbox/Pages/FolderListViewPage.xaml
huynhsontung: The spacing looks good

image

PR #357, opened 2024-04-14
chore: add tooltips to overflowed textblocks

United600:
Example of overflow text on a GridView Item
Example of overflow text on a ListView Item
Example of overflow text on a ListView Item #2

### DIFF
--- a/Screenbox/Controls/MediaListViewItem.xaml
+++ b/Screenbox/Controls/MediaListViewItem.xaml
@@ -52,47 +52,48 @@
             ToolTipService.ToolTip="{strings:Resources Key=IsPlaying}"
             Visibility="Collapsed" />
 
-        <Grid Grid.Column="1">
-            <TextBlock
-                x:Name="TrackNumberText"
-                Padding="0,0,8,0"
-                HorizontalAlignment="Right"
-                VerticalAlignment="Center"
-                Foreground="{ThemeResource TextFillColorTertiaryBrush}"
-                IsHitTestVisible="False"
-                Style="{StaticResource CaptionTextBlockStyle}"
-                Text="{Binding TrackNumberText, FallbackValue=''}"
-                TextWrapping="NoWrap"
-                Visibility="{x:Bind IsTrackNumberVisible}" />
-            <FontIcon
-                x:Name="ItemMediaTypeIcon"
-                HorizontalAlignment="Center"
-                VerticalAlignment="Center"
-                Glyph="{Binding Converter={StaticResource MediaGlyphConverter}}"
-                IsHitTestVisible="False"
-                Visibility="{x:Bind IsIconVisible}" />
-            <Button
-                x:Name="PlayButton"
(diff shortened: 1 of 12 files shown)


Screenbox/Pages/PlayerPage.xaml
```diff
@@ -259,9 +259,11 @@
 
                 <TextBlock
                     x:Name="SubtitleText"
+                    Foreground="{StaticResource TextFillColorSecondaryBrush}"
                     OpticalMarginAlignment="TrimSideBearings"
+                    Style="{StaticResource BodyTextBlockStyle}"
                     Text="{x:Bind ViewModel.Media.AltCaption, Mode=OneWay, FallbackValue={}}"
-                    TextTrimming="CharacterEllipsis"
+                    TextWrapping="NoWrap"
```
huynhsontung, line 266: Is there any reason why only the subtitle text has text trimming and not the title text?

Screenbox/Controls/MediaListViewItem.xaml
```diff
@@ -52,63 +52,60 @@
...
+                <Button.Resources>
+                    <StaticResource x:Key="SubtleButtonBackground" ResourceKey="HyperlinkButtonBackground" />
+                    <StaticResource x:Key="SubtleButtonBackgroundPointerOver" ResourceKey="HyperlinkButtonBackgroundPointerOver" />
+                    <StaticResource x:Key="SubtleButtonBackgroundPressed" ResourceKey="HyperlinkButtonBackgroundPressed" />
+                    <StaticResource x:Key="SubtleButtonForeground" ResourceKey="AccentTextFillColorPrimaryBrush" />
+                    <StaticResource x:Key="SubtleButtonForegroundPointerOver" ResourceKey="HyperlinkButtonForegroundPointerOver" />
+                    <StaticResource x:Key="SubtleButtonForegroundPressed" ResourceKey="HyperlinkButtonForegroundPressed" />
+                </Button.Resources>
+            </Button>
+        </Grid>
```
huynhsontung: Adding an extra layer of Grid may negatively affect the performance, especially when this control is intended to be used in a ListView. I'm not sure how much it affects but my rule of thumb has always been to keep the XAML tree as flat as possible.

huynhsontung (commented):
Thank you for using the behavior in more places. It makes a lot of sense.

Screenbox/Pages/PlayerPage.xaml
United600: Both don't need it, doesn't this commit removes it? I'm confused sorry.

Screenbox/Controls/MediaListViewItem.xaml
United600: I agree, keeping the tree as flat as possible should be a priority, do I remove it?

huynhsontung: Yes, please

PR #389, opened 2024-06-16
style: hide multi instance setting card on Xbox

United600:


huynhsontung:
Yes. Good call. Thank you!

PR #393, opened 2024-06-20
feat: add sort options for music pages

### LINKED ISSUE #370: Various sorting options for albums/artists, etc.
I think I'm not the only one who prefers a compact view of the media library, not based on the newfangled sorting by first letters. To be honest, I hate it... It’s convenient for me to navigate by the labels. For example, sort albums by date added. This makes it easier to remember what is where. It is often impossible to remember the name of the album... and when you come to scroll through a long list.
 There is such a function in the Windows music player and you can attach it. Unfortunately, even there it is impossible to make sure that when you click on the “Music” item, it opens the albums and not the tracks in the described order. (Although now I thought that it would be more logical for the tracks to open, since this is music) So the second point is at your discretion.
  As an idea, you can place the search by letters (the letters themselves) on the left, rather than separating them.

huynhsontung:
Add options to sort the music pages. Sort options for each page:
- Songs: title, album, artist, released year
- Albums: title, artist, released year
- Aritsts: N/A

Closes #370

### DIFF
--- a/Screenbox/App.xaml
+++ b/Screenbox/App.xaml
@@ -124,7 +124,6 @@
                             <TextBlock
                                 Style="{StaticResource SubtitleTextBlockStyle}"
                                 Text="{Binding Group.Key}"
-                                TextTrimming="CharacterEllipsis"
                                 TextWrapping="NoWrap" />
                         </Button>
                     </DataTemplate>
--- a/Screenbox/App.xaml
+++ b/Screenbox/App.xaml
@@ -116,11 +116,16 @@
                         <Button
                             MinWidth="{StaticResource GridViewHeaderItemMinHeight}"
                             MinHeight="{StaticResource GridViewHeaderItemMinHeight}"
+                            HorizontalAlignment="Stretch"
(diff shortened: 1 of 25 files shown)


Screenbox/App.xaml
```diff
@@ -116,11 +116,16 @@
...
+                            HorizontalAlignment="Stretch"
                             Background="Transparent"
                             BorderBrush="Transparent"
                             ClickMode="Hover"
                             IsEnabled="{Binding Group.Count, Converter={StaticResource DoubleToBoolConverter}, Mode=OneWay}">
-                            <TextBlock Style="{StaticResource SubtitleTextBlockStyle}" Text="{Binding Group.Key}" />
+                            <TextBlock
+                                Style="{StaticResource SubtitleTextBlockStyle}"
+                                Text="{Binding Group.Key}"
+                                TextTrimming="CharacterEllipsis"
```
United600: SubtitleTextBlockStyle style already sets the `TextTrimming="CharacterEllipsis"`

PR #395, opened 2024-06-30
feat: track media date added to the library

huynhsontung:
Track the date a song or an album was added to the library and let the user sort using it.

image


nero2077:
I just installed the program and pointed it to my music folder but the program didn't sort them correctly 
It's like instead of checking when the music file was added to my computer it checks when it was added to the program and I think it syncs the folder in chunks so it didn't sort them correctly  

This is how my music library looks like in Windows Media Player
image

And this is how my music folder loos like when it is sorted by date
image

Here how it looks in Screenbox
image

PR #403, opened 2024-07-13
feat: Add option to not display controls on pause

vessel00:
I often use screenshot software (e.g. pixpin, snipaste) to take a screenshot of the video directly and copy it to the clipboard, which is very easy to share. Therefore I seldom need to save the frame as a file. However, the controls keep displaying when paused, which is not convenient for taking screenshots directly. So I added a setting to control whether the controls will be displayed when paused. The controls will disappear after a while when paused. When pause with spacebar, it will not show the controls.


huynhsontung:
LGTM! Rebased to resolve a conflict.

PR #434, opened 2024-08-27
fix: unintended UI flipping in RTL languages

### LINKED ISSUE #429: Unintended UI flipping in right-to-left languages
The image is displayed upside down, here is a screenshot in Windows Media Player
image
And here is a picture from Screenbox
image
As I wrote here:
https://github.com/huynhsontung/Screenbox/issues/418#issuecomment-2309680883
The problems are that the application is configured in Windows in Hebrew mode and the operating system creates RTL for everything possible, including what is not needed...
In my opinion, if possible, you should set the application as English, and only the text in it will be in Hebrew.
If not possible, write to me and I will try to investigate the issue!

I use Windows 11, the latest version of the app released in the store.

huynhsontung:
Enforce LTR and RTL flow in specific UI elements. Fixes #429 

image

image

image

image

### DIFF
--- a/Screenbox/Controls/AcceleratorService.cs
+++ b/Screenbox/Controls/AcceleratorService.cs
@@ -29,7 +29,8 @@ namespace Screenbox.Controls
             {
                 string shortcut = ToShortcut(accelerator);
                 ToolTipService.SetToolTip(element,
-                    string.IsNullOrEmpty(shortcut) ? value : $"{value} ({shortcut})");
+                    string.IsNullOrEmpty(shortcut) ? value :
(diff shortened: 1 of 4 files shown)


Y-PLONI:
What about Windows Store?
Do you have a way to enforce LTR or vice versa - RTL?
I don't know what the problem is, but right now the text in the Windows Store is displayed incorrectly.
I started to reverse it in Crowdin but it's a bit complicated...

huynhsontung:
@Y-PLONI Does it still look reversed on the Store? It looks correct from my Store dashboard. Please avoid reversing the text in Crowdin. The translation should be independent of how the text should be displayed.

image

Y-PLONI:
Looks great! Thank you very much!
I took care of the translation, now it is in the true and correct order.

PR #454, opened 2024-09-28
l10n: Added Sinhala Translation

sipsuru:
Completed Trsnalation for si-LK


huynhsontung:
Thank you for your translation!

PR #459, opened 2024-10-03
chore: clean up solution and project files

United600:
- Remove AnyCPU and ARM platforms
- Sync Core platform min version with main project
- Remove ResXManager XLIFF setting


huynhsontung:
I wouldn't touch the project files at the moment. I will look into migrating these to dotnet 9, which will convert these projects to SDK style.

PR #466, opened 2024-10-15
style: adjust icon vertical alignment in buttons

United600:
Based on the design pattern outlined in the Figma WinUI document, the icon should be vertically centered within buttons.

Button with text and icon

The margin also makes sure that buttons maintain a consistent height in the icon only layout, especially important for those that hide the text content in the minimal visual state.

### DIFF
--- a/Screenbox/Styles/Button.xaml
+++ b/Screenbox/Styles/Button.xaml
@@ -55,6 +55,9 @@
     <!--  Uniform padding to ensure the text baseline aligns with TextBlocks (Caption) that have centered vertical alignment  -->
     <Thickness x:Key="HyperlinkButtonUniformPadding">11,6.75,11,6.75</Thickness>
 
+    <!--  Margin to vertically center an icon within a button with default padding, ensures the height remains consistent even when it contains only an icon  -->
+    <Thickness x:Key="IconButtonStandardIconMargin">0,1.95,0,0.9</Thickness>
+
     <Thickness x:Key="SplitButtonPadding">12,6,12,7.05</Thickness>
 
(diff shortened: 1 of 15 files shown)


Screenbox/Styles/Button.xaml
```diff
@@ -55,6 +55,9 @@
     <!--  Uniform padding to ensure the text baseline aligns with TextBlocks (Caption) that have centered vertical alignment  -->
     <Thickness x:Key="HyperlinkButtonUniformPadding">11,6.75,11,6.75</Thickness>
 
+    <!--  Margin to vertically center an icon within a button with default padding, ensures the height remains consistent even when it contains only an icon  -->
+    <Thickness x:Key="IconButtonStandardIconMargin">0,1.95,0,0.9</Thickness>
```
huynhsontung, line 59: If it's the standard margin, we should set an implicit style that applies this margin on all FontIcon in the app, instead of making individual changes in each XAML file.

United600: All? We want to restrict it to only those within a button, or do you have something like this in mind:

```xml
<Style x:Key="ButtonStandardIconFontIconStyle" TargetType="FontIcon">
    <Setter Property="FontSize" Value="{StaticResource StandardIconFontSize}" />
    <Setter Property="Margin" Value="0,1.95,0,0.9" />
</Style>
```

I thought about it but what if we use any other type of icon (SymbolIcon, ImageIcon, etc.)?

United600: Perhaps including "Standard" in the name was not the best choice. While I've only tested with 16 font size icons as per the spec, it should center icons of any other font size (Assuming default button padding or comparable uneven padding)

PR #473, opened 2024-10-23
fix: correct ImageBrush orientation for RTL languages

### LINKED ISSUE #429: Unintended UI flipping in right-to-left languages

United600:
Resolves #429 regression on some images. `ImageBrush`, unlike the [Image](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.image?view=winrt-26100#flowdirection-for-image) control, doesn't enforce a `LeftToRight` direction.

### DIFF
--- a/Screenbox/Controls/CommonGridViewItem.xaml
+++ b/Screenbox/Controls/CommonGridViewItem.xaml
@@ -19,23 +19,24 @@
         <Grid
             Grid.Row="0"
             Height="{x:Bind ThumbnailHeight, Mode=OneWay}"
-            Background="{ThemeResource SolidBackgroundFillColorSecondaryBrush}"
             BorderBrush="{ThemeResource ControlElevationBorderBrush}"
             BorderThickness="1"
-            CornerRadius="{x:Bind CornerRadius}">
-            <IconSourceElement
-                x:Name="PlaceholderIcon"
-                HorizontalAlignment="Center"
-                VerticalAlignment="Center"
-                Foreground="{ThemeResource TextFillColorSecondaryBrush}"
-                IconSource="{x:Bind PlaceholderIconSource, Mode=OneWay}"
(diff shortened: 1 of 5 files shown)


Screenbox/Controls/CommonGridViewItem.xaml
```diff
@@ -19,23 +19,24 @@
...
-                x:Name="PlaceholderIcon"
-                HorizontalAlignment="Center"
-                VerticalAlignment="Center"
-                Foreground="{ThemeResource TextFillColorSecondaryBrush}"
-                IconSource="{x:Bind PlaceholderIconSource, Mode=OneWay}"
-                Visibility="{x:Bind ThumbnailSource, Converter={StaticResource InverseEmptyObjectToVisibilityConverter}, Mode=OneWay}" />
-            <Border CornerRadius="{x:Bind CornerRadius}">
-                <Image
+            CornerRadius="{x:Bind CornerRadius}"
+            FlowDirection="LeftToRight">
```
huynhsontung, line 25: Wouldn't this force LeftToRight to the child element as well? In this case, wouldn't the icon be flipped?

United600: > Wouldn't this force LeftToRight to the child element as well? In this case, wouldn't the icon be flipped?

Yes. No, there's no change in the icons direction as they are already "forced" to LeftToRight.

PR #506, opened 2024-12-14
feat: Migrate to dotnet 9

huynhsontung:
Not functional due to incompatible dependencies. WinUI and CommunityToolkit is not yet supported. LibVlcSharp needs to be updated to not include WindowsAppSDK.


huynhsontung:
Superseded by #767

PR #522, opened 2024-12-28
chore: add media type disabled icons

United600:
- Adds audio off and movies off glyph
- Adds new icon `ResourceDictionary`
- Changes more icons to use our custom font family
- Changes `PlaylistView` selection command bar remove glyph to match the one used on flyouts


huynhsontung:
Just to confirm, the new icon dictionary is just for custom icon resources, right? Fluent icons are still left as-is?

United600:
> Just to confirm, the new icon dictionary is just for custom icon resources, right? Fluent icons are still left as-is?

Yes, at this point, I've only added the custom icon resources, since the code points (some) are harder to check compared to the system symbol font.
Do you want to add every icon in use?

huynhsontung:
No, that won't be necessary. I just want to clarify the intention for the font dictionary.

PR #530, opened 2025-01-05
feat: add option to always resume from the last position

### LINKED ISSUE #495: Always continue playback from last position
I wish to be able to always resume playback at where I left without having to manually click a button every time.

Having an option in the settings to enable always resume where the media stopped 

**Additional context**

Request from a Discord user

> *Talking about --qt-continue=2 arg*
> this qt command is supposed to continue the videos where I left them without asking.
> So I wonder if there is a libvlc command that does the same job.

huynhsontung:
Closes #495

Add the ability to always resume from the last position when opening a previously played file. This PR also refactor `LastPositionTracker` to participate in dependency injection. The instance is now hosted by the `SeekBarViewModel` instead of the `PlayerPageViewModel`.

### DIFF
--- a/Screenbox/Pages/SettingsPage.xaml
+++ b/Screenbox/Pages/SettingsPage.xaml
@@ -318,7 +318,8 @@
                 <ctc:SettingsCard
                     Margin="{StaticResource SettingsCardMargin}"
                     Description="{strings:Resources Key=SettingsRestorePlaybackPositionDescription}"
-                    Header="{strings:Resources Key=SettingsRestorePlaybackPositionHeader}">
+                    Header="{strings:Resources Key=SettingsRestorePlaybackPositionHeader}"
+                    HeaderIcon="{ui:FontIcon Glyph=&#xEF3B;}">
                     <ToggleSwitch IsOn="{x:Bind ViewModel.RestorePlaybackPosition, Mode=TwoWay}" />
                 </ctc:SettingsCard>
 
--- a/Screenbox/Pages/SettingsPage.xaml
+++ b/Screenbox/Pages/SettingsPage.xaml
@@ -315,6 +315,13 @@
                     <ToggleSwitch AutomationProperties.HelpText="{x:Bind SettingsEnqueueAllFilesInFolderCard.Description}" IsOn="{x:Bind ViewModel.EnqueueAllFilesInFolder, Mode=TwoWay}" />
                 </ctc:SettingsCard>
 
+                <ctc:SettingsCard
+                    Margin="{StaticResource SettingsCardMargin}"
+                    Description="Continue playing from where you last stopped when opening a file"
+                    Header="Always resume from last position">
+                    <ToggleSwitch IsOn="{x:Bind ViewModel.RestorePlaybackPosition, Mode=TwoWay}" />
+                </ctc:SettingsCard>
+
                 <!--  Player section  -->
                 <TextBlock Style="{StaticResource SettingsSectionHeaderTextBlockStyle}" Text="{strings:Resources Key=SettingsCategoryPlayer}" />
 
--- a/Screenbox/Pages/SettingsPage.xaml
+++ b/Screenbox/Pages/SettingsPage.xaml
@@ -172,7 +172,7 @@
                     </interactivity:Interaction.Behaviors>
                     <Button
                         x:Name="AddMusicFolderButton"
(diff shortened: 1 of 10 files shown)


Screenbox/Pages/SettingsPage.xaml
```diff
@@ -315,6 +315,13 @@
                     <ToggleSwitch AutomationProperties.HelpText="{x:Bind SettingsEnqueueAllFilesInFolderCard.Description}" IsOn="{x:Bind ViewModel.EnqueueAllFilesInFolder, Mode=TwoWay}" />
                 </ctc:SettingsCard>
 
+                <ctc:SettingsCard
+                    Margin="{StaticResource SettingsCardMargin}"
+                    Description="{strings:Resources Key=SettingsRestorePlaybackPositionDescription}"
+                    Header="{strings:Resources Key=SettingsRestorePlaybackPositionHeader}">
```
United600: What do you think about using the replay icon?

```suggestion
                    Header="{strings:Resources Key=SettingsRestorePlaybackPositionHeader}"
                    HeaderIcon="{ui:FontIcon Glyph=&#xEF3B;}">
```

Screenbox/Pages/SettingsPage.xaml
```diff
@@ -315,6 +315,13 @@
...
+                    <ToggleSwitch IsOn="{x:Bind ViewModel.RestorePlaybackPosition, Mode=TwoWay}" />
```
United600: When using ToggleSwitch, ComboBox, Slider, etc., as content we should bind AutomationProperties.HelpText to the SettingsCard/Expander description.

United600:
Does it make sense to add a progress bar to the CommonGridViewItem? Perhaps it should be limited to the recents page.

Grid view item with a progress bar

Screenbox/Pages/SettingsPage.xaml
huynhsontung: That's not an ideal icon for this option, but it should be good enough for now. We should consider what to do when we have more specific options that we can't find an icon for. Currently, when options get too specific, we group them into a SettingsExpander and give the expander an icon, like the Gestures options. Is this a good enough solution or do we need something else?

huynhsontung:
> Does it make sense to add a progress bar to the CommonGridViewItem? Perhaps it should be limited to the recents page.

@United600 I think that's a great feature. It should be applied to all grid views, so adding it to the CommonGridViewItem seems reasonable.

United600:
> > Does it make sense to add a progress bar to the CommonGridViewItem? Perhaps it should be limited to the recents page.
> 
> @United600 I think that's a great feature. It should be applied to all grid views, so adding it to the CommonGridViewItem seems reasonable.

Yeah but after thinking about it a bit more, I don't think it should be visible on audio items. Well... we do support audiobooks and other long-form audio content. Nevermind.

Screenbox/Pages/SettingsPage.xaml
United600: I think it effectively conveys the feature. The anti-clockwise arrow is a good starting point and base, what else should we add to improve it?

For the rest, we should first explore the fluent icons document for inspiration. If we can't find a suitable icon, we should look for general ideas and adapt them to "our" style

huynhsontung: I think the idea of "playback position" is lost in this icon specifically. However, it's not a blocker.

For the rest, I understand that we can custom-craft our own icons for each option. The question is: Should we? Do we really need each option to have its own icon? If you dig deep enough in the Windows Settings, you will see pages with a mix of icon and no-icon options. I think we should embrace no-icon options or hide them somehow.

> Currently, when options get too specific, we group them into a SettingsExpander and give the expander an icon, like the Gestures options.

This is a way to hide no-icon options.

United600: > I think the idea of "playback position" is lost in this icon specifically. However, it's not a blocker.

I can see that, resume vs. replay.

> For the rest, I understand that we can custom-craft our own icons for each option. The question is: Should we? Do we really need each option to have its own icon?

For top-level options an icon should be required.

> If you dig deep enough in the Windows Settings, you will see pages with a mix of icon and no-icon options. I think we should embrace no-icon options or hide them somehow.

I agree, certain options within a subpage (maybe even a group) may not require an icon.

PR #533, opened 2025-01-06
style: improve command bar style and resources

United600:
- Optimizes and improves the custom `AppBarButton` styles
- Adds a `CommandBar` extension that enables some customization of the MoreButton
- Adds a custom style and animation for the MoreButton
- Reverses the selection `CommandBar` translation animation so it enters from the top and exits through the bottom
- Changes the PlayQueue select all shortcut to function as a toggle, switching between selecting all and deselecting all items
- Adds access keys to the media Details `CommandBar`
- Improves the gamepad navigation of the media Details and FolderView pages


Screenbox/App.xaml
```diff
@@ -84,15 +84,17 @@
                     <CornerRadius x:Key="CircularCornerRadius">99</CornerRadius>
 
                     <!--  Animations  -->
-                    <ctAnimations:ImplicitAnimationSet x:Key="SimpleShowAnimations">
+                    <ctAnimations:ImplicitAnimationSet x:Key="FadeInImplicitAnimation">
                         <ctAnimations:OpacityAnimation
                             EasingType="Linear"
+                            From="0"
```
huynhsontung: From values are intentionally not defined to avoid pop-in effect when chaining animations. Say, for example, a fade-out animation is playing and interrupted by a fade-in animation. Instead of continuing from a half fade-out state, the fade-in animation will set the opacity to 0 before transitioning to 1. Although I have not personally encountered this pop-in effect but, that was my thought process

Screenbox/Controls/Extensions/CommandBarExtensions.cs
```diff
@@ -0,0 +1,122 @@
...
+
+/// <summary>
+/// Provides attached dependency properties for the <see cref="CommandBar"/> control.
+/// </summary>
+internal class CommandBarExtensions
+{
+    /// <summary>
+    /// Attached <see cref="DependencyProperty"/> for binding a <see cref="KeyboardAccelerator"/> to the more <see cref="Button"/> of the associated <see cref="CommandBar"/>
+    /// </summary>
+    public static readonly DependencyProperty MoreButtonKeyboardAcceleratorsProperty = DependencyProperty.RegisterAttached(
```
huynhsontung: Why is KeyboardAccelerators plural here? It's setting only one accelerator

huynhsontung: Also, this doesn't seem to be used anywhere. 🤔

huynhsontung (commented):
Just some questions about the CommandBarExtensions. Do we need to add more keyboard accelerators for the more button? What's the use case for feature?

United600:
> Just some questions about the CommandBarExtensions. Do we need to add more keyboard accelerators for the more button? What's the use case for feature?

Not really. It's more of a leftover.

Screenbox/App.xaml
United600: Without it the opacity animation wouldn't play, only reason I added it.

Screenbox/Controls/Extensions/CommandBarExtensions.cs
United600: > Why is KeyboardAccelerators plural here? It's setting only one accelerator

This allows setting multiple `KeyboardAccelerator`, like this:
```xml
<extensions:CommandBarExtensions.MoreButtonKeyboardAccelerators>
    <KeyboardAccelerator Key="A" />
    <KeyboardAccelerator Key="C" />
</extensions:CommandBarExtensions.MoreButtonKeyboardAccelerators>
```

Screenbox/App.xaml
United600: Ah! I guess that one doesn't need it.
It's more of a precaution after I had issues with the one for the more button.

Screenbox/Controls/Extensions/CommandBarExtensions.cs
United600: I guess not that's not entirely accurate. That was the goal, but after some light [reading](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.uielement.keyboardaccelerators?view=winrt-26100), it would need to be an IList<>.

huynhsontung:
Please remove the unused method to avoid confusion. The keyboard accelerator setter is not technically accurate anyway.

huynhsontung (approved):

PR #542, opened 2025-01-19
feat: add customizable application theme

SKure27:
Added a drop down menu on the settings page which allows you to switch the app theme between light, dark, and auto (default).

This is my first time using visual studio as well as working on a UWP app. I could see there was a lot of generated code but wasn't sure what the official way to add to the ui was so I just made changes manually.

It would be nice if the feature could be copied into the actual app (just implemented better then how I did it).


huynhsontung (approved):
I went ahead and refactored your changes to better align with our structure.

image

PR #544, opened 2025-01-25
style: critical error panel redesign

United600:
Changes the design of the critical error panel to resemble the BSOD.

Edit: Mockup added
Page with critical error grid

### DIFF
--- a/Screenbox/Pages/MainPage.xaml
+++ b/Screenbox/Pages/MainPage.xaml
@@ -356,7 +356,6 @@
                 HorizontalScrollMode="Disabled">
                 <StackPanel>
                     <TextBlock
-                        x:Name="CriticalErrorDescription"
                         MaxWidth="768"
                         HorizontalAlignment="Left"
                         d:Text="{strings:Resources Key=CriticalErrorDirect3D11NotAvailable}"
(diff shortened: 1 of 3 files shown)


huynhsontung (approved):
Thank you for this change. I'm just adding a screenshot for some visual reference :)

image

Screenbox/Pages/MainPage.xaml
```diff
@@ -333,29 +333,81 @@
...
+                            <Grid.RowDefinitions>
+                                <RowDefinition Height="Auto" />
+                                <RowDefinition Height="*" />
+                            </Grid.RowDefinitions>
+
+                            <TextBlock
+                                AutomationProperties.AccessibilityView="Content"
+                                FontWeight="SemiLight"
+                                Style="{StaticResource SubtitleTextBlockStyle}">
+                                <Run Text="For more information about this issue and possible fixes, visit" /><LineBreak />
```
huynhsontung: Are we not localizing this string?

United600: Oh absolutely. IIRC I chose not to do it because I felt it could be improved rather than making a direct copy of the BSOD.
Does the "possible fixes" part makes sense in our case? Maybe include instructions to create an issue?

PR #550, opened 2025-02-02
style: improve split button design and interaction

United600:
- Splits the previous behavior into an extension and a behavior
- Improvements to the behavior, focus visual, and XY navigation

Old|New
-|-
Old focus visual behvior|New focus visual behavior

- Adds a new split button style that better adheres to the design guidelines
- Adds a accent split button style
- Adds a thickness converter


Screenbox/Converters/ThicknessFilterConverter.cs
```diff
@@ -0,0 +1,75 @@
...
+{
+    public enum ThicknessFilterKind { None, Left, Top, Right, Bottom, LeftRight, TopBottom };
+
+    /// <summary>
+    /// Identifies the <see cref="FilterProperty"/> property.
+    /// </summary>
+    public static readonly DependencyProperty FilterProperty = DependencyProperty.Register(
+        nameof(Filter), typeof(ThicknessFilterKind), typeof(ThicknessFilterConverter), new PropertyMetadata(null));
+
+    public ThicknessFilterKind Filter { get; set; } = ThicknessFilterKind.None;
```
huynhsontung, line 24: I think the property name `Filter` is confusing. This converter converts from Thickness to Thickness with some edge zero out. The name makes me think it is filtering four values of Thickness into a single number. Maybe `Exclude` is a more descriptive name.

United600: To me it conveys the objective pretty clearly (filters out and alters a value), and it aligns with the WinUI terminology (the [inspiration](https://github.com/marcelwgn/microsoft-ui-xaml/blob/dev/splitbutton-dark-theme-border-fix/dev/SplitButton/SplitButton.xaml) and the [corner radius](https://github.com/microsoft/microsoft-ui-xaml/blob/winui2/main/dev/CommonStyles/CornerRadius_themeresources.xaml) converter)

huynhsontung (approved):

PR #556, opened 2025-02-07
feat: improve and add new shortcuts

United600:
- adds missing keyboard resources to the .csproj
- adds VLC-like Save Snapshot shortcut (`Shift` + `S`)
- adds YouTube-like Next button shortcut (`Shift` + `N`)
- adds YouTube-like Previous button shortcut (`Shift` + `P`)
- fixes some unwanted search shortcut behaviors
- removes the selection row opacity to fix the focus visual visibility
- changes the toggle player visibility controller shortcut to use the `View` button, freeing up the `Y` button to be used for search, as per the documentation [here](https://learn.microsoft.com/en-us/windows/apps/design/devices/designing-for-tv#search-experience) and [here](https://learn.microsoft.com/en-us/windows/apps/design/input/gamepad-and-remote-interactions#accelerator-support)


Screenbox/Screenbox.csproj
```diff
@@ -461,52 +461,76 @@
       <LastGenOutput>Resources.generated.cs</LastGenOutput>
       <SubType>Designer</SubType>
     </PRIResource>
+    <PRIResource Include="Strings\ar-SA\KeyboardResources.resw" />
```
huynhsontung, line 464: Oh we were missing the keyboard resources huh

United600: 🙄 My bad, I thought that it was automatic with the Crowdin action

huynhsontung: I just realized, what's stopping people from not following the keyboard constraints? For example, you can enter more than one character for a resource that requires exactly one capital letter.

United600: I think [Crowdin](https://support.crowdin.com/string-management/#string-editing) provides some assistance there.

United600: > I just realized, what's stopping people from not following the keyboard constraints? For example, you can enter more than one character for a resource that requires exactly one capital letter.

So, Crowdin does have some [syntax](https://support.crowdin.com/expression-syntax-elements/) support, but I don't think [.resw](https://store.crowdin.com/resw) supports it.
If I'm reading this [right](https://support.crowdin.com/string-management/#adding-strings) correctly, only formats that allow editing of the source files are supported (eg. [resx](https://store.crowdin.com/resx)). Maybe it's only a limitation on adding and editing strings, but it still provides validation.

huynhsontung: Without Crowdin, we have no way of enforcing these rules, right? Many contributors prefer creating a PR and editing the .resw files directly. This is not a major concern since Crowdin still has precedence over the resource files; it's just something to keep in mind.

Issue #361, opened 2024-04-19
Allow forcing dark/light theme


RibShark:
**Is your feature request related to a problem? Please describe.**
I tend to use a light theme on my system, but for certain programs such as the terminal and media players, I prefer a dark theme; most media players support changing this from the system default.

**Describe the solution you'd like**
I would like the ability to change the theme away from the system default and to my preferred theme.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #561, opened 2025-02-10
feat: improve app theme change

### LINKED ISSUE #539: Option to Select Theme
I generally prefer a light theme on windows, however, for specific apps like media players I prefer dark mode. 

Would it be possible to add the option to switch between light and dark mode?

My only other option would be to switch to a different app. I really enjoy using this one and don't want to switch but this is kind of a deal breaker for me when choosing a media player.

**Additional context**
N/A

huynhsontung:
Closes #539 #361

Move the theme change setting to its own category on the settings page. Changing the theme on the settings page takes effect immediately without requiring an app restart.

I also fixed an issue with the `PlayerPage` theme propagation, where the page would not respect the theme requested by the parent.

https://github.com/user-attachments/assets/3eb10a55-53d9-42bc-b57d-4d08db4bb152

### DIFF
--- a/Screenbox/Strings/en-US/Resources.resw
+++ b/Screenbox/Strings/en-US/Resources.resw
@@ -893,19 +893,22 @@
   <data name="SettingsEnqueueAllFilesInFolderHeader" xml:space="preserve">
     <value>Add all files in folder to queue</value>
   </data>
-  <data name="Auto" xml:space="preserve">
-    <value>Auto</value>
+  <data name="ThemeAuto" xml:space="preserve">
+    <value>Use system setting</value>
   </data>
-  <data name="Dark" xml:space="preserve">
+  <data name="ThemeDark" xml:space="preserve">
     <value>Dark</value>
   </data>
-  <data name="Light" xml:space="preserve">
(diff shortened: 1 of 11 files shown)


Screenbox/Strings/en-US/Resources.resw
```diff
@@ -893,19 +893,22 @@
...
+  <data name="ThemeDark" xml:space="preserve">
     <value>Dark</value>
   </data>
-  <data name="Light" xml:space="preserve">
+  <data name="ThemeLight" xml:space="preserve">
     <value>Light</value>
   </data>
   <data name="SettingsThemeSelectionDescription" xml:space="preserve">
-    <value>Select a theme to match your Windows settings or choose a custom one</value>
+    <value>Change the app's overall color scheme</value>
```
United600, line 906: I would use the string used by the inbox apps. Or at least remove the mention of "color".

```suggestion
    <value>Select which app theme to display</value>
```

huynhsontung: I don't think we should have "app theme" to be repeated. The description becomes redundant then.

PR #562, opened 2025-02-10
style: update system caption button colors

United600:
Changes the caption button colors according to the design specifications. Uses subtle brushes for the hover and pressed background colors.
Fixes the caption buttons colors in light mode with the "Show accent color on title bars and windows borders" setting enabled
Fixes the dialogs not respecting the user's selected theme

Old|New
-|-
Shows the hover color of the current minimize caption button|Shows the hover color of the new minimize caption button

### Task list

- [x] Fix the transition between video and audio content in light mode
- [x] Trigger the color change with the app theme setting


huynhsontung:
Is this PR ready to merge, btw? I see there are incomplete tasks on the task list.

United600:
> Is this PR ready to merge, btw? I see there are incomplete tasks on the task list.

Not really. AFAIK it's not regressing anything, but there's room for improvement. Maybe moving the logic to `App.xaml.cs` to consolidate both the main page and player page implementation? Do we need some sort of a theme helper?
I also believe addressing the in-app theme change is imperative.

Screenbox/Pages/MainPage.xaml.cs
```diff
@@ -54,6 +57,10 @@ public MainPage()
             //Register a handler for when the window changes focus
             Window.Current.CoreWindow.Activated += CoreWindow_Activated;
 
+            //Register a handler for when the theme mode changes
+            _themeListener = new ThemeListener();
+            _themeListener.ThemeChanged += ThemeListener_OnThemeChanged;
```
huynhsontung: There is an `ActualThemeChanged` event in the `FrameworkElement` class. I think this listener is a legacy code.

https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.frameworkelement.actualthemechanged?view=winrt-26100#windows-ui-xaml-frameworkelement-actualthemechanged

United600: Thanks, didn't knew about it.

Screenbox/Pages/MainPage.xaml.cs
```diff
@@ -131,12 +134,55 @@ public void NavigateContent(Type pageType, object? parameter)
...
+        }
+
+        private void UpdateSystemCaptionButton()
+        {
+            var titleBar = ApplicationView.GetForCurrentView().TitleBar;
+            if (titleBar != null)
+            {
+                // Rest
+                var buttonBackgroundColor = (Color)Application.Current.Resources["CaptionButtonBackground"];
+                var buttonForegroundColor = (Color)Application.Current.Resources["CaptionButtonForeground"];
```
huynhsontung: Since the application theme never actually changes, you can't rely on application theme resources to get the color values for the active theme. If the system theme is Dark and the user changes the theme to Light in the app, these lines will still return values for Dark theme.

Screenbox/App.xaml
```diff
@@ -17,12 +17,49 @@
...
+                            <StaticResource x:Key="CaptionButtonBackgroundPressed" ResourceKey="SubtleFillColorTertiary" />
+                            <!--  Foreground colors ignore the alpha channel, so we flattened them by layering them over the SolidBackgroundFillColorBase and its associated subtle color  -->
+                            <!--  TextFillColorPrimary  -->
+                            <Color x:Key="CaptionButtonForeground">#FFFFFF</Color>
+                            <!--  TextFillColorPrimary  -->
+                            <Color x:Key="CaptionButtonForegroundPointerOver">#FFFFFF</Color>
+                            <!--  TextFillColorSecondary  -->
+                            <Color x:Key="CaptionButtonForegroundPressed">#D1D1D1</Color>
+                            <!--  TextFillColorDisabled  -->
+                            <Color x:Key="CaptionButtonForegroundInactive">#717171</Color>
```
huynhsontung: There is no way to explicitly get a theme-specific value from a theme resource, so you may need to convert these into static resources with names specific to the Dark and Light themes.

huynhsontung:
This PR is getting out of scope. Please keep the PR related to the caption button colors only.

United600:
> This PR is getting out of scope. Please keep the PR related to the caption button colors only.

Are you referring to changes in the content dialogs? During testing I noticed that it didn't followed the requested theme changes but you're right that it should be a separate PR.

huynhsontung:
I moved the player theme switching logic on the player page into code behind. The player page only switches to Dark theme active when the media is a video and the player is expanded.

United600:
> I moved the player theme switching logic on the player page into code behind. The player page only switches to Dark theme active when the media is a video and the player is expanded.

Nice, I tried something similar but had issues with the transition between video and audio playback.

PR #563, opened 2025-02-12
chore: minor editorconfig update

United600:
Some minor updates and, adds the `dotnet_sort_system_directives_first = true` and `csharp_indent_case_contents_when_block = false` configs.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,6 +4,8 @@ root = true
 # All files
 [*]
 indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
 
 # Solution files
 [*.sln]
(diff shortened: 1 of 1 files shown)


huynhsontung:
Setting `dotnet_sort_system_directives_first = true` now may make a lot of files need to be reformatted.

United600:
> Setting `dotnet_sort_system_directives_first = true` now may make a lot of files need to be reformatted.

True, if not almost all. I saw that it was the set on the [runtime](https://github.com/dotnet/runtime/blob/main/.editorconfig#L83) repository and that it is recommended.

huynhsontung:
The default value for this was `true` as well. Not sure how it was set to `false`. https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/dotnet-formatting-options#dotnet_sort_system_directives_first

United600:
> The default value for this was `true` as well. Not sure how it was set to `false`. https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/dotnet-formatting-options#dotnet_sort_system_directives_first
> 
> 
> 
> 

Maybe it's a relatively new setting that it isn't automatically applied on some projects (like Net Native)?

PR #584, opened 2025-03-26
chore: refactor and add comments to thickness converter

United600:
Enhancements to `ThicknessFilterConverter`:
* Updated the `ThicknessFilterConverter` class to be sealed and inherit from `DependencyObject`. Added detailed XML documentation and an example to illustrate usage.
* Modified the `ThicknessFilterKind` enum to support bitwise combinations, allowing more flexible filtering. 
* Renamed the `ExcludeConvert` method to `ExtractThickness` and updated its logic to handle bitwise filter combinations.


huynhsontung:
Oh XAML has native support for flags?

huynhsontung (approved):

United600:
> Oh XAML has native support for flags?

After seeing a shortcut modifier I thought there must be a way to replicate it, and then I remember you mentioning the flag enum attribute. I tried and it worked flawlessly.

PR #616, opened 2025-05-27
chore: minor improvements to the ThicknessFilterConverter

United600:
Refactors the converter for improved usability (specially for C#) and simplicity.

- Moves the enum outside the class for improved usability
- Renames the enum to follow the naming [guidelines](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-classes-structs-and-interfaces#naming-enumerations) by using the plural form
- Renames the `ExtractThickness` method to `Convert` for improved usability
- Simplifies the filtering logic by replacing the switch statement with the `HasFlag` method
- Small adjustments to the converter `XAML` resource names
- Improves and adds some comments remarks

### DIFF
--- a/Screenbox/App.xaml
+++ b/Screenbox/App.xaml
@@ -40,14 +40,14 @@
                     <!--<converters:ThicknessFilterConverter x:Key="BottomOnlyThicknessFilterConverter" Filter="Bottom" />-->
                     <!--<converters:ThicknessFilterConverter x:Key="LeftRightThicknessFilterConverter" Filter="Left,Right" />-->
                     <converters:ThicknessFilterConverter x:Key="TopBottomThicknessFilterConverter" Filter="Top,Bottom" />
-                    <!--<converters:ThicknessFilterConverter x:Key="TopLeftThicknessFilterConverter" Filter="Top,Left" />-->
+                    <!--<converters:ThicknessFilterConverter x:Key="TopLeftThicknessFilterConverter" Filter="Left,Top" />-->
                     <!--<converters:ThicknessFilterConverter x:Key="TopRightThicknessFilterConverter" Filter="Top,Right" />-->
-                    <!--<converters:ThicknessFilterConverter x:Key="BottomLeftThicknessFilterConverter" Filter="Bottom,Left" />-->
-                    <!--<converters:ThicknessFilterConverter x:Key="BottomRightThicknessFilterConverter" Filter="Bottom,Right" />-->
-                    <!--<converters:ThicknessFilterConverter x:Key="LeftRightTopThicknessFilterConverter" Filter="Left,Right,Top" />-->
-                    <!--<converters:ThicknessFilterConverter x:Key="LeftRightBottomThicknessFilterConverter" Filter="Left,Right,Bottom" />-->
-                    <converters:ThicknessFilterConverter x:Key="TopBottomLeftThicknessFilterConverter" Filter="Top,Bottom,Left" />
-                    <converters:ThicknessFilterConverter x:Key="TopBottomRightThicknessFilterConverter" Filter="Top,Bottom,Right" />
+                    <!--<converters:ThicknessFilterConverter x:Key="BottomLeftThicknessFilterConverter" Filter="Left,Bottom" />-->
+                    <!--<converters:ThicknessFilterConverter x:Key="BottomRightThicknessFilterConverter" Filter="Right,Bottom" />-->
+                    <!--<converters:ThicknessFilterConverter x:Key="LeftRightAndTopThicknessFilterConverter" Filter="Left,Top,Right" />-->
+                    <!--<converters:ThicknessFilterConverter x:Key="LeftRightAndBottomThicknessFilterConverter" Filter="Left,Right,Bottom" />-->
+                    <converters:ThicknessFilterConverter x:Key="TopBottomAndLeftThicknessFilterConverter" Filter="Left,Top,Bottom" />
+                    <converters:ThicknessFilterConverter x:Key="TopBottomAndRightThicknessFilterConverter" Filter="Top,Right,Bottom" />
                     <ctConverters:BoolNegationConverter x:Key="BoolNegationConverter" />
                     <ctConverters:BoolToVisibilityConverter x:Key="BoolToVisibilityConverter" />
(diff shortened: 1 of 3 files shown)


Screenbox/Converters/ThicknessFilterConverter.cs
```diff
@@ -7,179 +7,119 @@
...
-    /// Extracts, using filters, the specified fields from a <see cref="Thickness"/> struct.
+    /// Extracts the specified fields from a <see cref="Thickness"/> struct.
     /// </summary>
-    /// <param name="thickness">The <see cref="Thickness"/> to convert.</param>
-    /// <param name="filterKind">An <see cref="Enum"/> that defines the filter type.</param>
+    /// <param name="thickness">The source <see cref="Thickness"/> to convert.</param>
+    /// <param name="filterKind">An enumeration that specifies the filter type.</param>
     /// <returns>A <see cref="Thickness"/> with only the fields specified by the filter, while the rest are set to 0.</returns>
-    public Thickness ExtractThickness(Thickness thickness, ThicknessFilterKind filterKind)
+    public Thickness Convert(Thickness thickness, ThicknessFilterKinds filterKind)
```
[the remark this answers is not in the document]
United600: It was intentional to optimize C# usage, for example:
```cs
var thicknessConverter = new ThicknessFilterConverter();
var exampleThickness= new Thickness(1, 1, 1, 1);
var thicknessFilter = ThicknessFilterKinds.Left;

...BorderThickness = thicknessConverter.Convert(exampleThickness, thicknessFilter);
```
instead of
```cs
var thickness = new Thickness(1, 1, 1, 1);
var thicknessConverter = new ThicknessFilterConverter()
{
    Filter = ThicknessFilterKinds.Left
};

...BorderThickness = (Thickness)thicknessConverter.Convert(thickness, typeof(Thickness), null, null);
```

Screenbox/App.xaml
```diff
@@ -40,14 +40,14 @@
...
-                    <converters:ThicknessFilterConverter x:Key="TopBottomLeftThicknessFilterConverter" Filter="Top,Bottom,Left" />
-                    <converters:ThicknessFilterConverter x:Key="TopBottomRightThicknessFilterConverter" Filter="Top,Bottom,Right" />
+                    <!--<converters:ThicknessFilterConverter x:Key="BottomLeftThicknessFilterConverter" Filter="Left,Bottom" />-->
+                    <!--<converters:ThicknessFilterConverter x:Key="BottomRightThicknessFilterConverter" Filter="Right,Bottom" />-->
+                    <!--<converters:ThicknessFilterConverter x:Key="LeftRightAndTopThicknessFilterConverter" Filter="Left,Top,Right" />-->
+                    <!--<converters:ThicknessFilterConverter x:Key="LeftRightAndBottomThicknessFilterConverter" Filter="Left,Right,Bottom" />-->
+                    <converters:ThicknessFilterConverter x:Key="TopBottomAndLeftThicknessFilterConverter" Filter="Left,Top,Bottom" />
+                    <converters:ThicknessFilterConverter x:Key="TopBottomAndRightThicknessFilterConverter" Filter="Top,Right,Bottom" />
```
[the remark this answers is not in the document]
United600: Another deliberate decision to follow the Thickness field order only in the Filter property.

huynhsontung, line 50: Why do we have `And` in the resource key anyway? Isn't `TopBottomRightThicknessFilterConverter` clear enough?

United600: > Why do we have And in the resource key anyway?

The goal was to establish a clearer distinction. I also considered going with `VerticalAndLeft`/`HorizontalAndTop` or maybe `EdgesAndLeft`/`SidesAndTop`.

> Isn't TopBottomRightThicknessFilterConverter clear enough?

Yes and no, it was inspired by this WinUI [enum](https://github.com/microsoft/microsoft-ui-xaml/blob/v2.8.7/dev/Common/CornerRadiusToThicknessConverter.idl#L22-L36). I can revert it, don't really have a strong preference or opinion either way.

United600:
Do you think the Filter dependency property should be plural as well? I believe properties that take an enum with flags should, for example:
[Modifiers](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.input.keyboardaccelerator.modifiers?view=winrt-26100#windows-ui-xaml-input-keyboardaccelerator-modifiers)

huynhsontung:
@United600 
> Do you think the Filter dependency property should be plural as well? I believe properties that take an enum with flags should

Yes, I agree. This makes sense since the property can combine multiple values.

Screenbox/Converters/ThicknessFilterConverter.cs
```diff
@@ -7,179 +7,119 @@
...
 /// <summary>
-/// An <see cref="IValueConverter"/> that takes an existing <see cref="Thickness"/> struct and returns a new one,
-/// using filters to keep only the specified fields and setting all others to 0.
-/// <example>For example:
+/// Defines constants that specify the filter type for a <see cref="ThicknessFilterConverter"/> instance.
+/// <para>This enumeration supports a bitwise combination of its member values.</para>
+/// </summary>
+/// <remarks>This enumeration is used by the <see cref="ThicknessFilterConverter.Filter"/> property.</remarks>
+[Flags]
+public enum ThicknessFilterKinds
```
huynhsontung, line 15: The enum name shouldn't be plural, though.

United600: In general or just the "Filter" bit?

United600: I've finally found one system [flags enum](https://learn.microsoft.com/en-us/uwp/api/windows.devices.enumeration.devicepairingkinds?view=winrt-26100
) that shares the same name termination. Still haven't come across a single WinRT flags enum that deviates from the recommended guidelines:
> ✔️ DO use a plural type name for an enumeration with bit fields as values, also called flags enum.

I'm also thinking about renaming the converter to "ThicknessFiltersConverter" so it aligns with the dependency property, what do you think?

huynhsontung: That works for me!

PR #623, opened 2025-06-14
feat: add option to change the app's display language

### LINKED ISSUE #604: Editable media information and language setting
I’d love to see a few features added to Screenbox:

1. **Editable Media Info**  
It’s great that we can view media information, but it would be even better if we could edit it directly within Screenbox. VLC has this feature, and it's very useful.

2. **Custom Thumbnails**  
VLC also allows users to change the thumbnail of a video, although it's currently broken. I’d like to see a working version of this in Screenbox.

3. **Resolution Icons**  
The VLC nightly build displays small resolution icons (HD, 2K, 4K, etc.) on the top right corner of video thumbnails. This is a nice touch for quickly identifying video quality and would be a great addition.

4. **Language Settings**  
There’s currently no option to change the app's language. It seems to follow the system language, which can be frustrating. My system is in another language, but I prefer using all my apps in English. A manual language selection option would be much appreciated.

Thanks for the great work!

huynhsontung:
Partially resolved #604 

image

### DIFF
--- a/Screenbox/Pages/SettingsPage.xaml
+++ b/Screenbox/Pages/SettingsPage.xaml
@@ -10,6 +10,7 @@
     xmlns:ctBehaviors="using:CommunityToolkit.WinUI.Behaviors"
     xmlns:ctc="using:CommunityToolkit.WinUI.Controls"
     xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:globalization="using:Windows.Globalization"
     xmlns:helpers="using:Screenbox.Helpers"
     xmlns:interactions="using:Screenbox.Controls.Interactions"
     xmlns:interactivity="using:Microsoft.Xaml.Interactivity"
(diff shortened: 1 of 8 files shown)


Screenbox/Pages/SettingsPage.xaml
```diff
@@ -286,6 +288,23 @@
...
                     </ComboBox>
                 </ctc:SettingsCard>
+                <ctc:SettingsCard
+                    Margin="{StaticResource SettingsCardMargin}"
+                    Description="{strings:Resources Key=SettingsLanguageSelectionDescription}"
+                    Header="{strings:Resources Key=SettingsLanguageSelectionHeader}"
+                    HeaderIcon="{ui:FontIcon Glyph=&#xF2B7;}">
+                    <ComboBox
+                        x:Name="LanguageComboBox"
+                        MinWidth="200"
```
United600, line 298: Do we need to make the MinWidth larger than that the rest? I guess it's to keep the list from changing width while scrolling.

Screenbox/Pages/SettingsPage.xaml
```diff
@@ -286,6 +288,23 @@
...
+                        ItemsSource="{x:Bind ViewModel.AvailableLanguages}"
```
United600, line 300:
```suggestion
                        AutomationProperties.HelpText="{x:Bind strings:Resources.SettingsLanguageSelectionDescription}"
                        ItemsSource="{x:Bind ViewModel.AvailableLanguages}"
```

United600:
Maybe we could add (depends on the difficulty) an initial entry to the list that resets the language to the system default (Tag = "") and make it the default option. To maintain consistency with the other personalization option.

```
System default
English (US)
The remaining languages alphabetically
```

Screenbox/Pages/SettingsPage.xaml
huynhsontung: Oh no, if the ComboBox is smaller than one of the items, the list is offset to the side. It's weird. Making the ComboBox 200 pixels wide is wide enough to fit all the items.

huynhsontung:
@United600 That makes sense. I added a system language option and sorted the list. I don't think English (US) need to be the second option.

image

United600:
> @United600 That makes sense. I added a system language option and sorted the list. I don't think English (US) need to be the second option.
> 
> image

I believe it does make some sense. System default first, followed by the app's "default" language.

I'm conflicted about this order. Even though it doesn't have right-to-left and other languages alphabets peppered in, at least it was "truly" alphabetical. We could use the `DisplayName` but on my tests it only worked with English (US) and the system language.

Screenbox/Pages/SettingsPage.xaml
United600: Oh right, I'm already "used" to it. Change the language to pt-PT and open the theme option ComboBox and _voilá_.

PR #624, opened 2025-06-14
style: add ListView style and improve the extension

United600:
This PR introduces a shared ListView style to enforce (and simplify) consistent UI across multiple XAML files. Certain properties remain in the container style and haven't been migrated yet, as they're still pending performance evaluation.

- Moves the extension files from the `Controls` directory to the project root
- Updates the `ApplicationViewExtensions` header to include the appropriate copyright notice
- Adds and improves the `ListViewExtensions` attached properties
- Adds a `MediaListViewStyle` to standardize the style
- Fixes the visual feedback on the search pages `ListView`

### DIFF
--- a/Screenbox/Extensions/ListViewExtensions.cs
+++ b/Screenbox/Extensions/ListViewExtensions.cs
@@ -3,7 +3,7 @@ using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Controls.Primitives;
 
-namespace Screenbox.Controls.Extensions
+namespace Screenbox.Extensions
 {
     public static class ListViewExtensions
     {
--- a/Screenbox/Extensions/ListViewExtensions.cs
+++ b/Screenbox/Extensions/ListViewExtensions.cs
@@ -1,4 +1,6 @@
-﻿using CommunityToolkit.WinUI;
+﻿#nullable enable
+
+using CommunityToolkit.WinUI;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Controls.Primitives;
@@ -68,7 +70,7 @@ public static class ListViewExtensions
     /// Identifies the attached dependency property that specifies the minimum height of the <see cref="ListView"/> item.
     /// </summary>
     public static readonly DependencyProperty ItemMinHeightProperty = DependencyProperty.RegisterAttached(
-        "ItemMinHeight", typeof(double), typeof(ListViewExtensions), new PropertyMetadata(null, OnItemMinHeightPropertyChanged));
+        "ItemMinHeight", typeof(double), typeof(ListViewExtensions), new PropertyMetadata(default(double), OnItemMinHeightPropertyChanged));
 
     /// <summary>
     /// Gets the value of the ItemMinHeight attached property from the specified <see cref="ListViewBase"/>.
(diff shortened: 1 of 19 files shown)


Screenbox/Extensions/ListViewExtensions.cs
```diff
@@ -0,0 +1,256 @@
...
+                listViewBase.ContainerContentChanging += ItemMinHeightOnContainerContentChanging;
+                listViewBase.Unloaded += OnListViewBaseUnloaded;
+            }
+        }
+    }
+
+    private static void ItemMinHeightOnContainerContentChanging(ListViewBase sender, ContainerContentChangingEventArgs args)
+    {
+        if (args.Phase > 0 || args.InRecycleQueue) return;
+        if (!IsApiContract13Present && ItemMarginProperty != null)
```
United600: Maybe we should switch the operation sequence

Screenbox/Extensions/ListViewExtensions.cs
```diff
@@ -0,0 +1,271 @@
...
+        "ItemMinHeight", typeof(double), typeof(ListViewExtensions), new PropertyMetadata(null, OnItemMinHeightPropertyChanged));
+
+    /// <summary>
+    /// Gets the value of the ItemMinHeight attached property from the specified <see cref="ListViewBase"/>.
+    /// </summary>
+    /// <param name="element">The <see cref="ListViewBase"/> from which the property value is read.</param>
+    /// <returns>The current value of the ItemMinHeight attached property on the specified <see cref="ListViewBase"/>.</returns>
+    public static double GetItemMinHeight(ListViewBase element)
+    {
+        return (double)element.GetValue(ItemMinHeightProperty);
```
huynhsontung: Nitpick: I see the default value for `ItemMinHeight` property is null. But here we are doing an unsafe cast. If that property is ever be null, an exception will be thrown here.

United600: 🤔, I thought the (... != null) guarded against that?

https://github.com/United600/Screenbox/blob/2eccbb59e7127f44427e21bdc1e573e543a8b3d6/Screenbox/Extensions/ListViewExtensions.cs#L224

Wouldn't that also apply to the Thickness and CornerRadius as well?

huynhsontung: Oops I missed this reply. Generally, if the property has a value type (like a `double`), then the default value should be the default value of the type (e.g. `0d`). Unless there is an explicit need for a nullable value type, the property should not be initialized as `null`. If you need to null-check it for something, then you should declare the type as nullable (like `double?`). Value type and `null` should not mix.

Screenbox/Extensions/ListViewExtensions.cs
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    private static void OnItemCornerRadiusPropertyChanged(DependencyObject sender, DependencyPropertyChangedEventArgs args)
+    {
+        if (sender is ListViewBase listViewBase)
+        {
+            listViewBase.ContainerContentChanging -= ItemCornerRadiusOnContainerContentChanging;
+            listViewBase.Unloaded -= OnListViewBaseUnloaded;
+
+            if (ItemCornerRadiusProperty != null)
```
huynhsontung: Also I missed this, but these conditions don't make sense. These are doing null check against the DependencyProperty objects, not the nullable values. Should be `GetItemCornerRadius() != null` instead. Same for other places.

huynhsontung (approved):

huynhsontung:
The CI is broken again 😔

Screenbox/Extensions/ListViewExtensions.cs
United600: Oh, really?

When I first start dabbling with dependency properties, I noticed that they were being loaded even when not explicitly set in XAML. To fix that, and learn, I studied the Community Toolkit implementation (like [this](https://github.com/CommunityToolkit/Windows/blob/559a2a846a7f1dcd0415a6f3801e85213cb5fa09/components/Extensions/src/ListViewBase/ListViewExtensions.AlternateRows.cs#L74) for example).
However, looks like I've adopted it without fully understanding how it worked.

Out of curiosity, why not use `args.NewValue != null` instead?

huynhsontung: Oh Community toolkit also uses this pattern. I may be wrong then. Maybe there is an equality overload for DependencyProperty. Although Visual Studio didn't see any overload when I last checked.

huynhsontung: > Out of curiosity, why not use `args.NewValue != null` instead?

In this case, that could work too.

United600: > Oh Community toolkit also uses this pattern. I may be wrong then. Maybe there is an equality overload for DependencyProperty. Although Visual Studio didn't see any overload when I last checked.

In any case, we probably should also fix the other extensions.

PR #634, opened 2025-06-27
Add launch page settings

adrianghc:
Add a setting and surrounding infrastructure to let users pick which page should appear when they launch the app (except when launching via file). Any page besides the settings page is supported, including second-level pages like the albums or artists pages of the music library.

### DIFF
--- a/Screenbox.Core/Enums/SettingsOptions.cs
+++ b/Screenbox.Core/Enums/SettingsOptions.cs
@@ -20,3 +20,15 @@ public enum ThemeOption
     Light,
     Dark
 }
+
+public enum LaunchPageOption
+{
+    Home,
+    Songs,
+    Albums,
+    Artists,
+    VideoFolders,
+    AllVideos,
+    Network,
+    PlayQueue
+}
--- a/Screenbox/Pages/MainPage.xaml.cs
+++ b/Screenbox/Pages/MainPage.xaml.cs
@@ -7,7 +7,9 @@ using System.Linq;
 using System.Numerics;
 using CommunityToolkit.Mvvm.DependencyInjection;
 using Screenbox.Core;
+using Screenbox.Core.Services;
 using Screenbox.Core.ViewModels;
+using Screenbox.Helpers;
 using Sentry;
 using Windows.ApplicationModel.Core;
 using Windows.ApplicationModel.DataTransfer;
@@ -35,6 +37,8 @@ namespace Screenbox.Pages
         private MainPageViewModel ViewModel => (MainPageViewModel)DataContext;
 
         private readonly Dictionary<string, Type> _pages;
+
+        private ISettingsService Settings => Ioc.Default.GetRequiredService<ISettingsService>();
 
         public MainPage()
         {
@@ -145,7 +149,12 @@ namespace Screenbox.Pages
             if (!ViewModel.PlayerVisible)
             {
                 SetTitleBar();
-                NavView.SelectedItem = NavView.MenuItems[0];
+                var launchPage = Settings.LaunchPage;
+                var launchPageTag = launchPage.GetNavPageFirstLevel();
+                var navItem = NavView.MenuItems
+                    .OfType<muxc.NavigationViewItem>()
+                    .FirstOrDefault(item => item.Tag?.ToString() == launchPageTag);
+                NavView.SelectedItem = navItem ?? NavView.MenuItems[0];
                 _ = ViewModel.FetchLibraries();
             }
         }
--- a/Screenbox/Pages/MusicPage.xaml.cs
+++ b/Screenbox/Pages/MusicPage.xaml.cs
@@ -1,16 +1,19 @@
 ﻿#nullable enable
 
-using CommunityToolkit.Mvvm.DependencyInjection;
-using Screenbox.Core;
-using Screenbox.Core.ViewModels;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using CommunityToolkit.Mvvm.DependencyInjection;
+using Screenbox.Core;
+using Screenbox.Core.Services;
+using Screenbox.Core.ViewModels;
+using Screenbox.Helpers;
 using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Media.Animation;
 using Windows.UI.Xaml.Navigation;
 using NavigationView = Microsoft.UI.Xaml.Controls.NavigationView;
 using NavigationViewSelectionChangedEventArgs = Microsoft.UI.Xaml.Controls.NavigationViewSelectionChangedEventArgs;
+using muxc = Microsoft.UI.Xaml.Controls;
 
 // The Blank Page item template is documented at https://go.microsoft.com/fwlink/?LinkId=234238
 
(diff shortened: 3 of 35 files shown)


United600:
Nice change. I am not so sure about the subpages, we might want to consider adding only the navigation view items. That approach would allow us to reuse them for a eventual jumplist support.

Also you were only required to add the English string, as Crowdin manages the rest. Could you please revert those changes?

adrianghc:
I did consider adding only the navigation view items, but I found that I didn't like this approach because it effectively forces on the user an arbitrary opinion of the app about which subpage is more important. For me personally, the default songs view of the music page is not useful, and I always switch to the albums view. This is an arbitrary preference on my side, but so it would be for the app to only let me launch at the songs view. It would leave the idea of this feature, to let the user control where to launch the app, only half-baked. If you don't share the app's opinion on which subpage is the most important, you're left with an extra navigation step to do each time.

Finally, on the default Media Player app of Windows, the behavior is that while there is no setting like I'm implementing, the app does remember the last page you had open, down to the subpage level. I didn't want Screenbox to be less-than in this regard. :)

Regarding localization, should I just revert the changes to all `Resources.resw` files except for en-US? I'm a native German speaker so I already added de-DE translations, would it make sense to keep those rather than going through Crowdin for them?

United600:
> I did consider adding only the navigation view items, but I found that I didn't like this approach because it effectively forces on the user an arbitrary opinion of the app about which subpage is more important.

It’s cool, I was just throwing my opinion out there.

> Finally, on the default Media Player app of Windows, the behavior is that while there is no setting like I'm implementing, the app does remember the last page you had open, down to the subpage level. I didn't want Screenbox to be less-than in this regard. :)

We should do the same at some point, same for the jumplist support.

> Regarding localization, should I just revert the changes to all `Resources.resw` files except for en-US? I'm a native German speaker so I already added de-DE translations, would it make sense to keep those rather than going through Crowdin for them?

Yes, the de-DE translation should be reverted as well, and yeah it's a bit of a pain. Crowdin will overwrite any `Resources.resw` except for the en-US. Sorry about that.

adrianghc:
I have reverted all the translations. I suppose I can only add the de-DE translation back through Crowdin after this PR has been merged?

United600:
> I have reverted all the translations. I suppose I can only add the de-DE translation back through Crowdin after this PR has been merged?

Thanks. Yes, and if you don't have an account, or don't want to create one, we can translate it.

adrianghc:
@huynhsontung Would you like me to implement Copilot's suggested changes? Other usages of `ArgumentOutOfRangeException` in the code base do not set an error message so I didn't implement this either, but I can do it if you'd like.

United600:
> @huynhsontung Would you like me to implement Copilot's suggested changes? Other usages of `ArgumentOutOfRangeException` in the code base do not set an error message so I didn't implement this either, but I can do it if you'd like.

Can't speak for him but (IMHO) the `ArgumentOutOfRangeException` changes aren't strictly necessary, don't go out of your way to implement them.

huynhsontung:
Please ignore the Copilot comments and suggestions. I was just testing the feature.

I noticed that you also include subpage in the settings. I want to implement subpage persistence so when a nav page is selected it goes to the last subpage, as you noted with the Media Player. In the meantime, this setting will do the job.

Sorry for the delay in reviewing. I have been extremely busy.

adrianghc:
Hi @huynhsontung, any chance we could revisit this? I think the check failed only because it used the deprecated Windows Server image, which I saw you fixed in the meantime.

Screenbox/Pages/MainPage.xaml.cs
```diff
@@ -36,6 +38,8 @@ public sealed partial class MainPage : Page, IContentFrame
 
         private readonly Dictionary<string, Type> _pages;
 
+        private ISettingsService Settings => Ioc.Default.GetRequiredService<ISettingsService>();
```
huynhsontung, line 41: Getting the service directly from the control breaks the MVVM pattern. Ideally, we want only the ViewModels to have access to the services. Then the control will get the value from the ViewModel.

Screenbox/Pages/MusicPage.xaml.cs
```diff
@@ -55,7 +60,20 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
...
+                {
+                    libraryNavItem = (muxc.NavigationViewItem?)LibraryNavView.MenuItems[0];
+                }
+                else
+                {
+                    var launchPageTag = launchPage.GetNavPageSecondLevel();
+                    libraryNavItem = LibraryNavView.MenuItems
+                        .OfType<muxc.NavigationViewItem>()
+                        .FirstOrDefault(item => item.Tag?.ToString() == launchPageTag);
+                }
```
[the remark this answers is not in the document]
huynhsontung: > Calling GetNavPageSecondLevel() without checking if the launch page actually has a second level will cause an exception for Home, Network, and PlayQueue options.

The Copilot comment is valid. The methods `GetNavPageFirstLevel` and `GetNavPageSecondLevel` have an implicit dependency, making the logic fragile. Despite the name, we want to avoid exceptions when calling methods out of order.

huynhsontung, line 75: With this flow, the second-level page is set each time the user navigates to the parent page. Is this expected? The settings text suggests that this behaviour only applies to app launches. However, in reality, it is set on every page navigation.

Screenbox.Core/Enums/SettingsOptions.cs
```diff
@@ -20,3 +20,15 @@ public enum ThemeOption
...
+{
+    Home,
+    Songs,
+    Albums,
+    Artists,
+    VideoFolders,
+    AllVideos,
+    Network,
+    PlayQueue
+}
```
huynhsontung, line 34: Having the launch page name as an enum is also problematic in the future. If we add new page types to the app, we have to add the page at the bottom of the enum. The order of this enum is fixed once it is released. We cannot change this order (only append to it) because doing so will break the user's settings.

huynhsontung (commented):
Sorry for the very late review! 😓

After reviewing the code, this solution appears somewhat messy for the problem it aims to address. Especially having the settings page as an enum, once released, it will be tough to change in the future. It might be more effective to implement a way to persist the Frame history across sessions, so that the user sees the last opened page when the app is launched. We probably can persist the navigation state using `Frame.GetNavigationState` and `Frame.SetNavigationState`.

Screenbox.Core/Enums/SettingsOptions.cs
adrianghc: This is true, but we could also hardcode the enum values, then we could insert new ones in the middle. Though that would not be all that elegant either.

adrianghc:
I understand your concerns. So, if I understand you right, you'd rather scrap this implementation and go with an entirely different approach, correct? In that case I would close this. I might be able to get around to the approach you propose if I find the time and then I can create a separate PR for that.

Screenbox/Pages/MusicPage.xaml.cs
adrianghc: From my testing, it actually only applies to app launches, not every page navigation.

PR #649, opened 2025-08-05
chore: refactor timing offset properties

United600:


huynhsontung:
There is actually a reason I didn't want to include the timing offset properties in the `IMediaPlayer` interface. I want `IMediaPlayer` to track the UWP [MediaPlayer](https://learn.microsoft.com/en-us/uwp/api/windows.media.playback.mediaplayer) class as much as possible. This way, we can be sure that any player backend we implement can follow the UWP standard. This is the assumption that enables FFmpeg integration. (https://github.com/huynhsontung/Screenbox/pull/333)

United600:
> There is actually a reason I didn't want to include the timing offset properties in the `IMediaPlayer` interface. I want `IMediaPlayer` to track the UWP [MediaPlayer](https://learn.microsoft.com/en-us/uwp/api/windows.media.playback.mediaplayer) class as much as possible. This way, we can be sure that any player backend we implement can follow the UWP standard. This is the assumption that enables FFmpeg integration. (#333)

Makes sense. `MediaPlayerElement` also I guess.

Screenbox.Core/ViewModels/PlayerControlsViewModel.cs
```diff
@@ -32,44 +32,13 @@ public sealed partial class PlayerControlsViewModel : ObservableRecipient,
 
         public bool ShouldBeAdaptive => !IsCompact && SystemInformation.IsDesktop;
 
-        public double SubtitleTimingOffset
-        {
-            // Special access. Consider promote to proper IMediaPlayer property
-            get => _subtitleTimingOffset = (_mediaPlayer as VlcMediaPlayer)?.VlcPlayer.SpuDelay / 1000 ?? 0;
```
huynhsontung, line 38: Changing this to an ObservableProperty lost the ability to get the value from VLC. Do the new properties not update at all?

huynhsontung: Yea I think the existing code doesn't even notify when the properties change. I think we are safe.

United600: Is this something the events added in the `VlcMediaPlayer.cs` can get it, or am I completely off?

huynhsontung: No, I think there is no event to update these properties from the VlcMediaPlayer so I think this change is safe. You are good.

PR #655, opened 2025-08-13
fix: InfoBar shadow artifact

### LINKED ISSUE #654: Small White Dot in Settings Page
**Describe the bug**

There is a small white triangle with shadow on the top right corner in settings page. Looks like a play button but got misplaced and cropped? Clicking on it does not do anything.

**How To Reproduce**

1. Open App
2. Go to Settings Page
3. See the upper right corner

**Expected behavior**

The corner should not have anything.

**Screenshots**

Image

**Environment**

 - OS: Windows 11 Dev 26200 25H2
 - Device: Surface Pro 10
 - App Version : 0.16.0.0

**Additional context**

United600:
Fixes #654. InfoBar doesn't play nice with ThemeShadow, so I removed it entirely. Alternatively, we could apply the ThemeShadow property only when `InfoBar.IsOpen="True"`.

Before|After
-|-
Home page with the InfoBar shadow clipping|Home page without the InfoBar shadow clipping
Settings page with the InfoBar shadow clipping|Settings page without the InfoBar shadow

### DIFF
--- a/Screenbox/Controls/NotificationView.xaml.cs
+++ b/Screenbox/Controls/NotificationView.xaml.cs
@@ -17,7 +17,6 @@ namespace Screenbox.Controls
         {
             this.InitializeComponent();
             DataContext = Ioc.Default.GetRequiredService<NotificationViewModel>();
-            InfoBar.Translation = new Vector3(0, 0, 16);
         }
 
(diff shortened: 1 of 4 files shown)


huynhsontung (approved):
Unfortunate that InfoBar doesn't play nice with ThemeShadow :(
LGTM

PR #665, opened 2025-09-03
chore: consolidate NavigationView keyboard and gamepad handling

United600:
Moves the NavigationView keyboard interaction features from the behavior to the custom control.
Initially, I considered converting it to an extension, but since the ListViewExtension PR moves the folder, I consolidated it instead.

- Adds dependency properties to `CustomNavigationView` for access keys, keyboard accelerators and styles for the back, close, pane toggle, pane search, and settings buttons
- Fixes gamepad support for the search button, which previously only worked in expanded display mode
- Improves the Overlay layout, removing the need to handle it on code
- Improves the Overlay with light-dismiss UI support (Xbox only)
- Adds new animations to the NavigationView sidebar and content
- Removes obsolete `NavigationViewKeyboardBehavior`

P.S. We should consider renaming the control to something like `NavigationViewEx`.

### DIFF
--- a/Screenbox/Controls/NavigationViewEx.cs
+++ b/Screenbox/Controls/NavigationViewEx.cs
@@ -68,6 +68,8 @@ public sealed partial class NavigationViewEx : NavigationView
 
     private const string ShadowCaster = "ShadowCaster";
 
+    private const string ContentGridFinalValue = "400";
+
     private static readonly ImplicitAnimationSet _slowFadeInAnimationSet = new()
     {
        new OpacityAnimation { To = 1, Duration = TimeSpan.FromMilliseconds(250), EasingType = EasingType.Linear }
@@ -80,16 +82,8 @@ public sealed partial class NavigationViewEx : NavigationView
     {
        new TranslationAnimation { From = "-48,0,0", To = "0,0,0", Duration = TimeSpan.FromMilliseconds(167), EasingMode = EasingMode.EaseOut },
     };
-    private static readonly ImplicitAnimationSet _contentShowAnimationSet = new()
-    {
-        new OpacityAnimation { To = 1, Duration = TimeSpan.FromMilliseconds(250), EasingType = EasingType.Linear },
-        new TranslationAnimation { To = "0,0,0", Duration = TimeSpan.FromMilliseconds(400), EasingMode = EasingMode.EaseOut }
-    };
-    private static readonly ImplicitAnimationSet _contentHideAnimationSet = new()
-    {
-        new OpacityAnimation { To = 0, Duration = TimeSpan.FromMilliseconds(167), EasingType = EasingType.Linear },
-        new TranslationAnimation { To = "0,-400,0", Duration = TimeSpan.FromMilliseconds(250), EasingMode = EasingMode.EaseIn }
-    };
+    private ImplicitAnimationSet? _contentShowAnimationSet;
+    private ImplicitAnimationSet? _contentHideAnimationSet;
 
     private Grid? _overlayRoot;
     private Border? _overlayChildBorder;
@@ -234,8 +228,7 @@ public sealed partial class NavigationViewEx : NavigationView
             _contentGrid = contentGrid;
 
             // Set implicit animations to play when ContentVisibility changes.
-            Implicit.SetShowAnimations(contentGrid, _contentShowAnimationSet);
-            Implicit.SetHideAnimations(contentGrid, _contentHideAnimationSet);
+            UpdateContentAnimations();
         }
 
         if (GetTemplateChild(ShadowCaster) is Grid shadowCaster)
@@ -411,7 +404,7 @@ public sealed partial class NavigationViewEx : NavigationView
     {
         if (_overlayRoot == null) return;
 
-        // Applies layout logic consistent with other SplitView content grids.
+        // Aligns the overlay layout with the SplitView content area.
         if (ContentVisibility == Visibility.Collapsed)
         {
             Grid.SetColumn(_overlayRoot, 0);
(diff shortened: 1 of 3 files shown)


Screenbox/Controls/NavigationViewEx.cs
```diff
@@ -1,207 +1,466 @@
...
+    }
+
+    private void OnLoaded(object sender, RoutedEventArgs e)
+    {
+        if (_splitView?.FindDescendant<Grid>() is { } splitViewGrid)
+        {
+            splitViewGrid.Children.Add(_overlayRoot);
+        }
+
+        if (_splitView?.FindDescendant<Border>(b => b.Name == "ContentBackground") is { } contentBackground)
```
United600: I can't find this control? What is supposed to do?

Screenbox/Controls/CustomNavigationView.cs
```diff
@@ -1,207 +1,466 @@
...
+            if (BackButtonStyle != null)
             {
-                new OpacityAnimation
-                {
-                    Duration = TimeSpan.FromSeconds(0.3),
-                    From = 1,
-                    To = 0
-                }
-            };
+                _closeButton.Style = BackButtonStyle;
```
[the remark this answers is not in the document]
United600: No, it’s intentional since it’s already set up that way.

United600:
We should add an ObservableProperty (or improve the PlayerVisible) to track player visibility, which can be bound to the Frame. This would reduce the visual tree when the play queue is empty. Also it should improve the app cold launch.

```diff
<controls:CustomNavigationView.Overlay>
-    <Frame x:Name="PlayerFrame" />
+    <Frame x:Name="PlayerFrame" Visibility="{x:Bind ViewModel.PlayerVisible, Mode=OneWay}" />
</controls:CustomNavigationView.Overlay>
```

Screenbox/Controls/NavigationViewEx.cs
huynhsontung: That is probably a regression. It was a custom element in the CustomNavigationView template and was supposed to be the background of the content pane. Setting the background of that `Border` helps blend the nav bar transition, as demonstrated in the video in https://github.com/huynhsontung/Screenbox/pull/189. I don't remember when the `ContentBackground` was removed. 🤔

huynhsontung: I found it. It was the custom SplitView template https://github.com/huynhsontung/Screenbox/pull/341. I removed it :)
 
I think code related to `ContentBackground` is safe to delete.

United600: Oh yeah, I can see what you’re saying about the background blend effect in the nav bar transition. Maybe we can investigate how to bring it back, unless it requires adding a custom SplitView style.

Screenbox/Controls/NavigationViewEx.cs
```diff
@@ -0,0 +1,526 @@
...
+        }
+    }
+
+    private void UpdateCloseButtonStyle()
+    {
+        if (_closeButton != null)
+        {
+            if (BackButtonStyle != null)
+            {
+                _closeButton.Style = BackButtonStyle;
```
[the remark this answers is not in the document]
United600: By design, they use the same style by default.

Screenbox/Controls/NavigationViewEx.cs
huynhsontung: The problem was the custom SplitView style doesn't follow the theme change. I think we can add the Border behind the content pane in code behind, without touching the SplitView template. We can listen to theme change events on the NavigationViewEx and change the border background accordingly.

United600: Or we could just "cheat" and simply remove the background entirely, like the system Settings app 😉.

https://github.com/user-attachments/assets/f7a4c2c9-9eef-4699-94c2-7ba200558ddf

United600: > The problem was the custom SplitView style doesn't follow the theme change. I think we can add the Border behind the content pane in code behind, without touching the SplitView template. We can listen to theme change events on the NavigationViewEx and change the border background accordingly.

Not sure about adding yet another background layer. Right now, I'm already darkening the light dismiss background (only on Xbox) just to fix the layer missing on Xbox. I could probably just always set the content light dismiss layer to collapsed.

huynhsontung: > Or we could just "cheat" and simply remove the background entirely, like the system Settings app 😉.

Honestly that may be the way. I remember I was against it for some reason. But it looks good now so why not.

United600: > Honestly that may be the way. I remember I was against it for some reason. But it looks good now so why not.

It doesn't look too good on the compact mode, probably why the system settings app skips it and switches straight to minimal mode.

huynhsontung: Anyway, we can deal with that in another PR. This PR looks good!

huynhsontung (approved):

Screenbox/Controls/NavigationViewEx.cs
United600: Agree. We can also deal with the Overlay visual tree visibility later.
Can't forget to remove the `ContentBackground` dead code.

United600:
The left translation animation is also cool, it gives a staggered feel to the it.

https://github.com/user-attachments/assets/972a0614-c62b-47c3-97bc-a8f0e1f4fd2c

Screenbox/Pages/MainPage.xaml.cs
```diff
@@ -51,8 +48,6 @@ public MainPage()
             // For example, when the app moves to a screen with a different DPI.
             coreTitleBar.LayoutMetricsChanged += CoreTitleBar_LayoutMetricsChanged;
 
-            NotificationView.Translation = new Vector3(0, 0, 16);
```
huynhsontung, line 54: Why did we remove the Translation for the Notification? This is necessary for ThemeShadow iirc

United600: Leftover of #655. It should've been removed there.

huynhsontung (approved):

PR #669, opened 2025-09-14
chore: migrates Newtonsoft.Json to System.Text.Json

United600:
Works perfectly fine on both Debug and Release builds, but I still have some concerns about the Store build. @rocksdanister Can you take a look and confirm that there are no regressions or hidden edge cases?

Followed the steps in this [guide](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/migrate-from-newtonsoft?pivots=dotnet-9-0).

### DIFF
--- a/Screenbox.Core/Helpers/WebView2Util.cs
+++ b/Screenbox.Core/Helpers/WebView2Util.cs
@@ -1,10 +1,10 @@
-﻿using Microsoft.UI.Xaml.Controls;
-using Microsoft.Web.WebView2.Core;
-using Newtonsoft.Json;
-using System;
+﻿using System;
 using System.IO;
 using System.Text;
+using System.Text.Json;
 using System.Threading.Tasks;
+using Microsoft.UI.Xaml.Controls;
+using Microsoft.Web.WebView2.Core;
 using Windows.System;
 
 namespace Screenbox.Core.Helpers;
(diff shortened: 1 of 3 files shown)


rocksdanister:
In Lively I am using [JsonConverter](https://github.com/rocksdanister/lively/blob/core-separation/src/Lively/Lively.Common/JsonConverters/LivelyControlModelConverter.cs) now.

Here that maybe overkill, don't have experience with System.Text.Json but `LivelyProperties.json` is designed with backward compatibility in mind - will not have required properties missing (unless the wallpaper itself is bad) and existing properties are not changed in future update.. so the checks are probably not necessary.

`LivelyInfo.json` can have missing fields depending on how and when its created but here we are deserializing so should be fine, its specification won't change in the future besides adding new properties.

United600:
Thanks for the info. It should work fine, but I’m still mildly concerned about .Net Native support for System.Text.Json.

rocksdanister:
You can verify by running with .NET Native in debug mode with this wallpaper (which I use for testing myself):
[Property API Test.zip](https://github.com/user-attachments/files/22347084/Property.API.Test.zip)

United600:
> You can verify by running with .NET Native in debug mode with this wallpaper (which I use for testing myself): [Property API Test.zip](https://github.com/user-attachments/files/22347084/Property.API.Test.zip)

The info looks correct but can't verify the properties. Debug and Release mode aren't really my concern.

LivelyWallpaper item info

rocksdanister:
> The info looks correct but can't verify the properties.

Just remove the music wallpaper check and run the wallpaper as visualizer.

https://github.com/huynhsontung/Screenbox/blob/3ac83ff51d1a45d3e27a19e24b559932ee4a38ad/Screenbox.Core/Services/LivelyWallpaperService.cs#L46

United600:
Correct again, but the app hangs after clicking `Continue` on the breakpoint (which I assume is expected).

https://github.com/user-attachments/assets/ed813d9c-e879-4439-8b47-1248f8b20d9a

PR #670, opened 2025-09-15
feat: customize delay before the controls automatically hide

### LINKED ISSUE #548: Hide/Minimize Playback Controls Options (progression bar etc...) and add color options for the progress bar and audio bar
 The fact that the playback options, progressions bar and video title stay on screen during a paused video and when moving the mouse once during the video disrupts the viewing experience. What would make this beautiful player perfect is if it would copy VLC in that regard: after moving the mouse during video, if the mouse stops moving, the video title, playback options and progressions bar hide themselves after about 1 second. Also if the video is paused, nothing stays on screen except a very quick appearance for 0,5 second of the "pause" symbol. Of course the best would be a setting option to chose the times one want for our selves. This way the focus is on the video, nothing of the viewing experience gets disrupted and if one pauses the video nothing stays on screen, which ensures a clear view on the video if one want to for example look at a beautiful frame without any UI elements.

Then the other thing are color options for the progress bar and audio bar, which are now in purple. I would really like to choose my own preferred color. Also the icons for minimizing immersive view and the playlist to the left side of the title in the upper left corner, I would like to be able to hide them as well if I want so only the title is showing at the top in the UI. 

With these small changes the player would be pretty much perfect and leagues above all other video players. 

**Additional context**

United600:
Combines the player controls settings into its own SettingsExpander and introduces an option to configure its hide delay (partially resolves #548).

Player options

There is definitely room for improvement. Not a fan of the Index switch statements, and I'd wish if we could use ReswPlus for the ComboBox values.


Screenbox.Core/ViewModels/PlayerPageViewModel.cs
```diff
@@ -576,9 +576,11 @@ public bool TryHideControls(bool skipFocusCheck = false)
...
         }
 
-        private void DelayHideControls(int delayInSeconds = 3)
+        private void DelayHideControls()
         {
             if (PlayerVisibility != PlayerVisibilityState.Visible || AudioOnly) return;
+            
+            int delayInSeconds = _settingsService.PlayerControlsHideDelay;
             _controlsVisibilityTimer.Debounce(() => TryHideControls(), TimeSpan.FromSeconds(delayInSeconds));
         }
```
United600:
```csharp
private void DelayHideControls(int extraSeconds = 0)
{
    if (PlayerVisibility != PlayerVisibilityState.Visible || AudioOnly) return;
     
    int delayInSeconds = _settingsService.PlayerControlsHideDelay + extraSeconds;
    _controlsVisibilityTimer.Debounce(() => TryHideControls(), TimeSpan.FromSeconds(delayInSeconds));
}
```

We could add a parameter to the `DelayHideControls()` so that we can increment the delay on the `FocusManagerOnFocusChanged()` like this:
```csharp
DelayHideControls(1);
```

United600:
Okay, I've streamlined the index switching logic and improved the ComboBox localization. Unfortunately, I had to add a converter (which is probably more complex than necessary) since I couldn’t get it working with a basic method like this on the ItemTemplate:

```cs
public string GetSecondsCountString(int value)
{
    return Strings.Resources.SecondsCount(value);
}
```

If you prefer using SelectedIndex, I can revert it.

Screenbox/Converters/IntToUnitStringConverter.cs
```diff
@@ -0,0 +1,76 @@
...
+/// Defines the unit types for formatting numeric values with localized and pluralized resource strings.
+/// </summary>
+/// <remarks>This enumeration is used by the <see cref="IntToUnitStringConverter.Unit"/> property.</remarks>
+public enum UnitType
+{
+    None = 0,
+    Albums = 1,
+    Songs = 2,
+    Seconds = 4,
+    Items = 11,
```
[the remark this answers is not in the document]
United600: Yeah, the gaps are there for time related units, like Millisecond, Minutes, Hours, etc.

Screenbox/Converters/IntToUnitStringConverter.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        return unit switch
+        {
+            UnitKind.Albums => Strings.Resources.AlbumsCount(value),
+            UnitKind.Songs => Strings.Resources.SongsCount(value),
+            UnitKind.Seconds => Strings.Resources.SecondsCount(value),
+            UnitKind.Items => Strings.Resources.ItemsCount(value),
+            _ => $"{value} {unit.ToString().ToLowerInvariant()}"
+        };
+    }
```
huynhsontung: I'm confused. How would we use this code? I only see the `seconds` unit being used.

huynhsontung:
> Okay, I've streamlined the index switching logic and improved the ComboBox localization. Unfortunately, I had to add a converter (which is probably more complex than necessary) since I couldn’t get it working with a basic method like this on the ItemTemplate:
> 
> ```cs
> public string GetSecondsCountString(int value)
> {
>     return Strings.Resources.SecondsCount(value);
> }
> ```
> 
> If you prefer using SelectedIndex, I can revert it.

I think you can probably do x:Bind to a static method inside the ItemTemplate, right? If not, a converter is fine.

Screenbox/Converters/IntToUnitStringConverter.cs
United600: It was made for the seconds unit, but I chose to expand it to make it future-proof. If you think it's unnecessary or detrimental, I can scale it back to its original focus.

United600:
> I think you can probably do x:Bind to a static method inside the ItemTemplate, right? If not, a converter is fine.

Tried, couldn't (probably a me problem).

Screenbox/Pages/SettingsPage.xaml
```diff
@@ -399,19 +400,33 @@
...
+                        </ctc:SettingsCard>
+                        <ctc:SettingsCard Header="{x:Bind strings:Resources.SettingsOnScreenControlsHideDelayHeader}">
+                            <ComboBox
+                                MinWidth="150"
+                                AutomationProperties.HelpText="{x:Bind strings:Resources.SettingsOnScreenControlsHideDelayHeader}"
+                                ItemsSource="{x:Bind ViewModel.PlayerControlsHideDelayOptions, Mode=OneWay}"
+                                SelectedItem="{x:Bind ViewModel.PlayerControlsHideDelay, Mode=TwoWay}">
+                                <ComboBox.ItemTemplate>
+                                    <DataTemplate x:DataType="x:Int32">
+                                        <TextBlock Text="{x:Bind Converter={StaticResource HideDelayOptionsToUnitStringConverter}}" />
```
huynhsontung: I found a way to bind directly to the `Resources` functions.

```suggestion
                                        <TextBlock Text="{x:Bind strings:Resources.SecondsCount((x:Int32))}" />
```

Reference: https://learn.microsoft.com/en-us/windows/apps/develop/platform/xaml/x-bind-markup-extension#pathless-casting

United600: Weird syntax, but works!

Well except that the ReswPlus pluralization is completly broken in certain languages, but that's for another pull request.
On-screen controls with the delay ComboBox opened

United600: Oh! If I add the "_Few" string is works, but shouldn't it return the "_Other" if it isn't present?

ReswPlus polish genertated code

huynhsontung: Something to debug 🤔 Definitely outside the scope of this PR though

Screenbox.Core/ViewModels/PlayerPageViewModel.cs
huynhsontung: I honestly forgot why I overrode the delay with the focus manager. Oh well, we will see I guess.

PR #678, opened 2025-09-24
fix: thickness converter possible null reference return

United600:
Minor refactor and improvement to the `ThicknessFiltersConverter` class and its related enum.

- Adds new `ThicknessFilterKinds` combinations
- Renames the `Convert` method to `Extract`
- Sets the `Extract` method to static for direct **XAML** usage
- Changes the `Extract` behavior, the `None` value now sets all fields to 0
- Improves the `Convert` method by returning `DependencyProperty.UnsetValue` instead of `null` for unsupported values
- Sets the `Filters` property default value back to `None`
- Removes the `partial` modifier, as it will no longer be needed for UWP modern NET migration

### DIFF
--- a/Screenbox/Converters/ThicknessFiltersConverter.cs
+++ b/Screenbox/Converters/ThicknessFiltersConverter.cs
@@ -39,9 +39,23 @@ public enum ThicknessFilterKinds
     /// </summary>
     Bottom = 8,
 
-    //Horizontal = Left | Right,
+    /// <summary>
+    /// Filters Left and Right values, sets Top and Bottom to 0.
+    /// </summary>
+    /// <remarks>This value combines the <see cref="Left"/> and <see cref="Right"/> flags.</remarks>
+    Horizontal = Left | Right,
 
-    //Vertical = Top | Bottom,
+    /// <summary>
+    /// Filters Top and Bottom values, sets Left and Right to 0.
+    /// </summary>
+    /// <remarks>This value combines the <see cref="Top"/> and <see cref="Bottom"/> flags.</remarks>
+    Vertical = Top | Bottom,
+
+    /// <summary>
+    /// Filters Left, Top, Right, and Bottom values.
+    /// </summary>
+    /// <remarks>This value combines the <see cref="Left"/>, <see cref="Top"/>, <see cref="Right"/>, and <see cref="Bottom"/> flags.</remarks>
+    All = Left | Top | Right | Bottom
 }
 
 /// <summary>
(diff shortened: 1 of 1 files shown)


Screenbox/Converters/ThicknessFiltersConverter.cs
```diff
@@ -108,25 +122,23 @@ public ThicknessFilterKinds Filters
...
-
-        if (filterKinds is not ThicknessFilterKinds.None)
+        if (filterKinds != ThicknessFilterKinds.None)
         {
             return new Thickness(
-                filterKinds.HasFlag(ThicknessFilterKinds.Left) ? result.Left : 0,
-                filterKinds.HasFlag(ThicknessFilterKinds.Top) ? result.Top : 0,
-                filterKinds.HasFlag(ThicknessFilterKinds.Right) ? result.Right : 0,
-                filterKinds.HasFlag(ThicknessFilterKinds.Bottom) ? result.Bottom : 0);
+                filterKinds.HasFlag(ThicknessFilterKinds.Left) ? thickness.Left : 0,
```
United600, line 135:
```suggestion
                filterKinds & ThicknessFilterKinds.Left) != 0 ? thickness.Left : 0,
```

Should we change to a bitwise operation? On modern .NET, it shouldn't matter but I'm not sure about .NET Native.

Screenbox/Converters/ThicknessFiltersConverter.cs
```diff
@@ -108,25 +122,23 @@ public ThicknessFilterKinds Filters
...
+                filterKinds.HasFlag(ThicknessFilterKinds.Top) ? thickness.Top : 0,
+                filterKinds.HasFlag(ThicknessFilterKinds.Right) ? thickness.Right : 0,
+                filterKinds.HasFlag(ThicknessFilterKinds.Bottom) ? thickness.Bottom : 0);
         }
 
-        return result;
+        return new Thickness();
```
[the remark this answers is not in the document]
United600: A empty `Thickness` has all values set to zero.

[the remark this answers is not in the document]
United600: Oh! The PR description was wrong.

Screenbox/Converters/ThicknessFiltersConverter.cs
huynhsontung: I don't think it matters on .NET Native either. I prefer the `HasFlag`  because it's clearer to me.

Screenbox/Converters/ThicknessFiltersConverter.cs
```diff
@@ -142,11 +154,10 @@ public object Convert(object value, Type targetType, object parameter, string la
     {
         if (value is Thickness thickness)
         {
-            var filtersType = Filters;
-            return Convert(thickness, filtersType);
+            return Extract(thickness, Filters);
         }
 
-        return null;
+        return DependencyProperty.UnsetValue;
```
huynhsontung, line 160: I'm not aware of the `DependencyProperty.UnsetValue`. Is there a functional difference between the two?

United600: Same here, I only learned about it from this toolkit [converter](https://github.com/CommunityToolkit/Windows/blob/559a2a846a7f1dcd0415a6f3801e85213cb5fa09/components/Converters/src/ColorToDisplayNameConverter.cs#L34). Not sure about the difference, we could just add `#nullable enable` to the file and call it a day. I'm not sure which one is the "correct".

PR #676, opened 2025-09-24
refactor: break up and refactor play queue components for clearer separation

huynhsontung:
- Refactored `MediaListViewModel` from a monolithic class into a clean MVVM architecture, separating business logic, domain models, and presentation layer for maintainability and testability.
- Implemented stateless services (`IPlaylistService`, `IPlaybackControlService`, `IMediaListFactory`) to encapsulate playlist operations, navigation logic, and media parsing, all injected via dependency injection.

PR #692, opened 2025-10-15
Feature: Enqueue in recent played

divyanshbhatia1:
I'm often frustrated when I resume watching a show from the Recently Played section, because Screenbox only plays that single file instead of continuing to the next episode in the same folder. This makes it inconvenient to binge or watch multiple episodes in sequence without manually returning to the folder each time.

When a user resumes playback from Recently Played, Screenbox should detect other playable files in the same folder and queue them automatically. After finishing the current file, playback should continue with the next file in order.
Optionally, there could be a setting like:

“Play next file in folder when resuming from Recently Played.”

Currently, the only workaround is to manually navigate to the folder and start playback from there to get sequential playback. This works, but it breaks the convenience of using the Recently Played shortcut.

**Additional context**
This feature would greatly improve the experience when watching TV series or multi-part videos, allowing smoother and more intuitive playback.

### DIFF
--- a/Screenbox.Core/ViewModels/HomePageViewModel.cs
+++ b/Screenbox.Core/ViewModels/HomePageViewModel.cs
@@ -177,7 +177,7 @@ namespace Screenbox.Core.ViewModels
         }
 
         [RelayCommand]
-        private void Play(MediaViewModel media)
+        private async Task Play(MediaViewModel media)
         {
             if (media.IsMediaActive)
             {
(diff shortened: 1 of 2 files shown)


United600:
Thanks for the contribution. We're currently in the middle of a major playlist refactoring, so this change will need to be put on hold for now. It may also require some adjustments after #676 merges.

TBH the `HomePage` recents list probably should also be refactored to group items like albums, artists, and playlists together, just like the Windows Media Player.

divyanshbhatia1:
Thanks for the update! Excited about the upcoming changes and will follow the pull request to adjust my changes accordingly. 
I also agree that the HomePage could use some refactoring and cleanup.

PR #693, opened 2025-10-15
fix: app theme combobox popup position

United600:
Under certain conditions, right-aligned `ComboBox` opens the popup selection in the wrong position microsoft/microsoft-ui-xaml#9567. As a workaround, setting the placeholder text to the longest item string resolves the issue.

I'm not sure how to handle the Language `ComboBox`. While we could set the placeholder to the longest item (ar-SA), it would make the popup unnecessarily wide for most users, since we aren't force installing all languages.

**Before:**

https://github.com/user-attachments/assets/cfa54228-41a0-4ff5-b2e5-c2db73549d3e

**After:**

https://github.com/user-attachments/assets/50e69939-19da-4cf4-8381-d38f950ba7dc


huynhsontung:
Nice catch! Thanks for referencing the WinUI issue

PR #695, opened 2025-10-17
feat: allow drop items into the middle of the play queue

huynhsontung:
Before

https://github.com/user-attachments/assets/af648a74-5fea-46d0-aff6-61e52386d256

After

https://github.com/user-attachments/assets/f94f906c-bc17-4351-a786-434d65d2fb1b

### DIFF
--- a/Screenbox/Extensions/ListViewExtensions.cs
+++ b/Screenbox/Extensions/ListViewExtensions.cs
@@ -1,5 +1,6 @@
 ﻿#nullable enable
 
+using System;
 using CommunityToolkit.WinUI;
 using Windows.Foundation;
 using Windows.UI.Xaml;
@@ -295,18 +296,23 @@ public static class ListViewExtensions
     {
         var position = e.GetPosition(listView);
 
-        foreach (var item in listView.Items)
+        foreach (var item in listView.ItemsPanelRoot.Children)
         {
-            var container = listView.ContainerFromItem(item) as ListViewItem;
+            var container = item as ListViewItem;
             if (container != null)
             {
                 var bounds = container.TransformToVisual(listView).TransformBounds(new Rect(0, 0, container.ActualWidth, container.ActualHeight));
 
                 // Check to see if the bounds of the item's container intersects the drop point
-                if (bounds.Top > position.Y)
+                if (bounds.Contains(position))
                 {
                     // Get the list view to tell us the index of the hit container
-                    return listView.IndexFromContainer(container);
+                    int dropIndex = listView.IndexFromContainer(container);
+
+                    // If the drop point is in the lower 40% of the container, we want to insert after it
+                    // The ratio is derived from WinUI 3 source code for ListViewBase drag and drop
+                    double threshold = bounds.Top + (bounds.Height * 0.6);
+                    return position.Y < threshold ? dropIndex : Math.Min(dropIndex + 1, listView.Items.Count);
                 }
             }
         }
--- a/Screenbox/Controls/PlayQueueControl.xaml.cs
+++ b/Screenbox/Controls/PlayQueueControl.xaml.cs
@@ -73,7 +73,7 @@ namespace Screenbox.Controls
             IReadOnlyList<IStorageItem>? items = await e.DataView.GetStorageItemsAsync();
             if (items?.Count > 0)
             {
-                int insertIndex = PlaylistListView.GetDropIndex(e) - 1;
+                int insertIndex = PlaylistListView.GetDropIndex(e);
                 await ViewModel.Playlist.EnqueueAsync(items, insertIndex);
             }
         }


United600:
Wow, I didn’t realize that drop events don't return an index. How does the reorder works internally?

I tried to extend the method to support `GridView`, but the `...WrapGrid` used as the `ItemsPanelTemplate` made it more complicated (plus it wasn't entirely correct).

```cs
public static int GetDropIndex(this ListViewBase listViewBase, DragEventArgs e)
{
    var position = e.GetPosition(listViewBase);
    bool isVertical = IsVerticalOrientation(listViewBase);

    foreach (var item in listViewBase.Items)
    {
        if (listViewBase.ContainerFromItem(item) is not SelectorItem container)
        {
            continue;
        }

        var bounds = container.TransformToVisual(listViewBase).TransformBounds(
            new Rect(0, 0, container.ActualWidth, container.ActualHeight));

        if (isVertical)
        {
            // Check to see if the bounds of the item's container intersects the drop point
            if (bounds.Top > position.Y)
            {
                // Get the list view to tell us the index of the hit container
                return listViewBase.IndexFromContainer(container);
            }
        }
        else
        {
            if (bounds.Left > position.X)
            {
                return listViewBase.IndexFromContainer(container);
            }
        }
    }
}

private static bool IsVerticalOrientation(ListViewBase listViewBase)
{
    var itemsPanel = listViewBase.ItemsPanelRoot;

    if (itemsPanel is ItemsStackPanel stackPanel)
    {
        return stackPanel.Orientation == Orientation.Vertical;
    }

    if (itemsPanel is ItemsWrapGrid wrapGrid)
    {
        return wrapGrid.Orientation == Orientation.Vertical;
    }

    if (itemsPanel is StackPanel panel)
    {
        return panel.Orientation == Orientation.Vertical;
    }

    //if (itemsPanel is VariableSizedWrapGrid variableSizedWrapGrid)
    //{
    //    return variableSizedWrapGrid.Orientation == Orientation.Vertical;
    //}

    if (itemsPanel is VirtualizingStackPanel virtualPanel)
    {
        return virtualPanel.Orientation == Orientation.Vertical;
    }

    return true;
}
```

Screenbox/Extensions/ListViewExtensions.cs
```diff
@@ -283,4 +284,36 @@ private static void ItemIsFocusEngagementEnabledOnContainerContentChanging(ListV
         if (args.Phase > 0 || args.InRecycleQueue || GetItemIsFocusEngagementEnabled(sender) is not { } isEnabled) return;
         args.ItemContainer.IsFocusEngagementEnabled = isEnabled;
     }
+
+    #region Extension Methods
```
United600, line 289: Regions? If you think the class is too long we could make it partial.

huynhsontung: It's not too long. The method I added is the only, by C# definition, actual extension method. The others are more of a XAML thing. They are both ListView extensions, just different types of extensions. Give it a region here for better grouping within the file.

huynhsontung:
> Wow, I didn’t realize that drop events don't return an index. How does the reorder works internally?

Oh man I checked the WinUI 3 source code, it's a whole process. The logic that goes into just detecting drag events and moving items for live updates is so involved. Essentially, they are still calculating the cursor position with respect to the item underneath it. So what we are doing is not that far off.

huynhsontung:
Let's extend this method to deal with `GridView` later when we have the usage for it. For now, it works well enough for `ListView`.

United600:
`// If the drop point is in the lower 40% of the container, we want to insert after it`
Is that true? Maybe only on WinUI 3? I'm not sure I like it.

Now it behaves differently from the reordering action. My, and user, expectation is that the item should always insert itself at the gap.

https://github.com/user-attachments/assets/a5e70ef9-333f-4a8c-ba39-c6fd0fe12d4c

huynhsontung:
Well you found an edge case. This is not the intended behavior. The 40% rule applies when you drop an item at the edge without triggering a reorder. Without the 40% rule, there will be cases where the item also doesn't go into the gap when you drop. The logic for live reordering is complicated. So it will be hard to 100% replicate their behavior. I will try again on my end.

United600:
> Well you found an edge case. This is not the intended behavior. The 40% rule applies when you drop an item at the edge without triggering a reorder. Without the 40% rule, there will be cases where the item also doesn't go into the gap when you drop. The logic for live reordering is complicated. So it will be hard to 100% replicate their behavior. I will try again on my end.

I believe you, I've just never experienced that 40% rule before. Don't waste time on it, it's infinitely better than it was.

I briefly looked [here](https://github.com/microsoft/microsoft-ui-xaml/blob/main/src/dxaml/xcp/dxaml/lib/ListViewBase_Partial_Reorder.cpp) but it's probably the wrong place, plus C++ 🫣

PR #696, opened 2025-10-18
fix: play button flicker when skipping items in play queue

huynhsontung:
Before:

https://github.com/user-attachments/assets/ab6369a6-12c7-426f-a830-17dfd79570a9

After:

https://github.com/user-attachments/assets/5b752242-96fa-47d0-a5a6-982ea5d30415

### DIFF
--- a/Screenbox.Core/ViewModels/PlayerControlsViewModel.cs
+++ b/Screenbox.Core/ViewModels/PlayerControlsViewModel.cs
@@ -251,7 +251,7 @@ namespace Screenbox.Core.ViewModels
         {
             _dispatcherQueue.TryEnqueue(() =>
             {
-                IsPlaying = sender.PlaybackState == MediaPlaybackState.Playing;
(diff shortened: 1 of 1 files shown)


United600:
😩 it's much more obvious on my system.

United600 (approved):

United600:
Should we start playing a item after skipping a paused item? This is our current approach, which differs from the Media Player, but it's not a general agreed behavior.

huynhsontung:
@United600 That's a good point. I was basing most of the player behaviour on YouTube. But I can see why the inbox Media Player makes sense. Plus, the logic is more straightforward with playing right after skipping lol. Loading the media to display the first frame then pause is messy with VLC.

United600:
> @United600 That's a good point. I was basing most of the player behaviour on YouTube. But I can see why the inbox Media Player makes sense. Plus, the logic is more straightforward with playing right after skipping lol. Loading the media to display the first frame then pause is messy with VLC.

Yeah. Like I said it's not general agreed behavior, actually most start playing.

PR #697, opened 2025-10-18
fix: rapid current item changing that break UI transitions

huynhsontung:
#676 introduced a regression where the current item is first changed to `null` when changing playlists. As a result, the current item can change from `item1` to `null` to `item2`, instead of going from `item1` directly to `item2`. This extra value change in rapid succession causes some UI logics to break.

### DIFF
--- a/Screenbox/Converters/MediaGlyphConverter.cs
+++ b/Screenbox/Converters/MediaGlyphConverter.cs
@@ -1,7 +1,7 @@
-﻿using Screenbox.Core.Enums;
+﻿using System;
+using Screenbox.Core.Enums;
 using Screenbox.Core.Helpers;
 using Screenbox.Core.ViewModels;
-using System;
 using Windows.Storage;
 using Windows.UI.Xaml.Data;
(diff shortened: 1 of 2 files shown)


United600:
Found another way to trigger the issue, and this time it wasn't a regression.

https://github.com/user-attachments/assets/83cde186-2489-4662-892d-2b5f6a2a2630

huynhsontung:
> Found another way to trigger the issue, and this time it wasn't a regression.

73312b9726288a7f319e24cdf2052797a59b2c88 should fix it now

United600:
> [73312b9](https://github.com/huynhsontung/Screenbox/commit/73312b9726288a7f319e24cdf2052797a59b2c88) should fix it now

I've never seen a stranger solution than this one 😂

huynhsontung:
It's a wonderful world of XAML 🌈

United600:
> It's a wonderful world of XAML 🌈

😩 I won't stand for XAML slander!

PR #700, opened 2025-10-18
feat: add settings privacy section

### LINKED ISSUE #771: Option to disabled "Resume where you left off" prompt
I am personally not a fan of the prompt to "Resume where you left off" and often prefer to just manually make use of the timeline/SeekBar.

Having a sub option to toggle `Track last position` and a sub option to `Clear tracked/last position history` below `Always resume from last position` would be pretty fancy.

N/A

**Additional context**
I understand this is a pedantic request; but, I wanted to throw the idea out there. I assume that not having the playback progress stored would result in the player acting like it does when a video is played for the first time, or is restarted after ending at 0:00 playback progress, thus effectively disabling the prompt.

United600:
- Added a new `Privacy` settings section
- Moved the `Show recent` card to the `Privacy` section
- Added a new `Remember playback position` setting
- Updated `FileService` to respect the `ShowRecent` setting

Closes #771

### DIFF
--- a/Screenbox/Pages/SettingsPage.xaml
+++ b/Screenbox/Pages/SettingsPage.xaml
@@ -28,6 +28,7 @@
     <Page.Resources>
         <!--<x:Double x:Key="SettingsCardActionIconMaxSize">14</x:Double>-->
         <Thickness x:Key="SettingsCardMargin">0,0,0,4</Thickness>
+        <x:Double x:Key="SettingsCardButtonMinWidth">120</x:Double>
 
         <converters:ResourceNameToResourceStringConverter x:Key="ResourceNameToResourceStringConverter" />
 
@@ -326,14 +327,6 @@
                     <ToggleSwitch AutomationProperties.HelpText="{strings:Resources Key=SettingsEnqueueAllFilesInFolderDescription}" IsOn="{x:Bind ViewModel.EnqueueAllFilesInFolder, Mode=TwoWay}" />
                 </ctc:SettingsCard>
 
-                <ctc:SettingsCard
-                    Margin="{StaticResource SettingsCardMargin}"
-                    Description="{strings:Resources Key=SettingsRestorePlaybackPositionDescription}"
-                    Header="{strings:Resources Key=SettingsRestorePlaybackPositionHeader}"
-                    HeaderIcon="{ui:FontIcon Glyph=&#xEF3B;}">
-                    <ToggleSwitch AutomationProperties.HelpText="{strings:Resources Key=SettingsRestorePlaybackPositionDescription}" IsOn="{x:Bind ViewModel.RestorePlaybackPosition, Mode=TwoWay}" />
-                </ctc:SettingsCard>
-
                 <!--  Player section  -->
                 <TextBlock Style="{StaticResource SettingsSectionHeaderTextBlockStyle}" Text="{strings:Resources Key=SettingsCategoryPlayer}" />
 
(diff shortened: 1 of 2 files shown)


United600:
We should also consider including the playback position history in the SettingsExpander items.

Remember playback positions SettingsExpander example

huynhsontung:
> We should also consider including the playback position history in the SettingsExpander items.

Disagree. If we want to show all the entries in the playback position history it should be outside of the SettingsExpander. Perhaps a popup. Anyway it's out of scope for this PR.

huynhsontung:
@United600 I made `LastPositionTracker` a singleton, so now SettingsPageViewModel can properly delete the saved progression.

United600:
> @United600 I made `LastPositionTracker` a singleton, so now SettingsPageViewModel can properly delete the saved progression.

Can confirm the cache is now being disposed of correctly.
I'll do a minor refactor based on #772, I like some of its choices.

huynhsontung:
May be unrelated, but what do you think of using the "progression" or "media progression" term instead of the "last position" term?

United600:
> May be unrelated, but what do you think of using the "progression" or "media progression" term instead of the "last position" term? 

Probably better.
Also the description can definitely be improved. It was inspired by some system setting but the first person feels weird.

United600:
I also wonder, should it be limited to video files? And is this last position "format" conducive to be added as a progress bar to the ‘CommonGridViewItem’?

United600:
Should I merge the resume and remember playback position cards? If so, which one do you prefer? No **InfoBar** is also an option.


https://github.com/user-attachments/assets/fb1d6a91-e458-4eaf-8591-6c5119f59d82

https://github.com/user-attachments/assets/0c2ebe05-fef4-45ed-b90d-8637e22802f8

huynhsontung:
If the "Always resume" option is under the "Playback position" expander, then I don't think we need to explain why the option is disabled. I thought the "Always resume" option would remain in the General section. This ordering works as well, and we don't need to deal with extra explanation.

huynhsontung:
> > May be unrelated, but what do you think of using the "progression" or "media progression" term instead of the "last position" term?
> 
> Probably better. Also the description can definitely be improved. It was inspired by some system setting but the first person feels weird.

Let's implement this in a separate PR.

huynhsontung:
> I also wonder, should it be limited to video files? And is this last position "format" conducive to be added as a progress bar to the ‘CommonGridViewItem’?

In most cases, I agree that this feature only makes sense with video files. But I think we should artificially limit this feature because it works in most cases. Perhaps we can make the "Always resume" option only affect video files. That option may break the music listening experience if turned on.

United600:
> If the "Always resume" option is under the "Playback position" expander, then I don't think we need to explain why the option is disabled. I thought the "Always resume" option would remain in the General section. This ordering works as well, and we don't need to deal with extra explanation.

Did some copywriting on both settings, what do you think?

imagem

United600:
> > I also wonder, should it be limited to video files? And is this last position "format" conducive to be added as a progress bar to the ‘CommonGridViewItem’?
> 
> In most cases, I agree that this feature only makes sense with video files. But I think we should artificially limit this feature because it works in most cases. Perhaps we can make the "Always resume" option only affect video files. That option may break the music listening experience if turned on.

Yeah, it does makes sense for audiobooks (and podcasts). Though our support for them isn't exactly stellar.

huynhsontung:
> > If the "Always resume" option is under the "Playback position" expander, then I don't think we need to explain why the option is disabled. I thought the "Always resume" option would remain in the General section. This ordering works as well, and we don't need to deal with extra explanation.
> 
> Did some copywriting on both settings, what do you think?

Just nitpicking: the description for "Playback position" is a bit wordy, and the first-person usage is... odd. Otherwise, it's good. I don't mind shipping it as-is.

United600:
> Just nitpicking: the description for "Playback position" is a bit wordy, and the first-person usage is... odd. Otherwise, it's good.

It was inspired by the search settings.

Search privacy settings

> I don't mind shipping it as-is.

I think the resume setting should be moved to the playback position expander items.

United600:
The 1424cdd commit moves and updates the "Always resume" setting. If the change doesn't feel right to you, I'll go ahead and revert it (or some parts).

PR #708, opened 2025-10-26
fix: player controls not appearing when the access key (Alt) is pressed

United600:
This pull request refactors how keyboard accelerator keys are defined and used in the player page, and improves control visibility logic related to key tips. The main changes include moving key definitions to a helper class for better maintainability, updating XAML bindings, and ensuring controls remain visible when key tips are shown.

**Keyboard Accelerator Key Refactoring:**

- Moves key definitions (such as Plus, Minus, Comma, and Period) from `PlayerPage.xaml.cs` to the `KeyboardAcceleratorHelper` class, making them reusable and easier to maintain
- Removes unused `pages` namespace import from `PlayerPage.xaml`, as key definitions are now in the helper.

**Control Visibility and Key Tips:**

- Adds a new observable property `KeyTipsVisible` to `PlayerPageViewModel`, and logic to ensure player controls are visible when key tips are visible
- Subscribes to `AccessKeyManager.IsDisplayModeEnabledChanged` in `PlayerPage.xaml.cs` to update the view model’s `KeyTipsVisible` property in response to access key display mode changes

**Other Improvements:**

- Improves flyout placement logic in `PlayerPage.xaml.cs`, including better support for right-to-left languages

### DIFF
--- a/Screenbox/Pages/PlayerPage.xaml
+++ b/Screenbox/Pages/PlayerPage.xaml
@@ -950,6 +950,7 @@
                     </VisualState.StateTriggers>
                     <VisualState.Setters>
                         <Setter Target="PlayQueueButtonIcon.Glyph" Value="{StaticResource PlaylistMusicMirroredGlyph}" />
+                        <Setter Target="PlayQueueFlyout.Placement" Value="BottomEdgeAlignedRight" />
                         <Setter Target="LivelyOptionsButtonIcon.Glyph" Value="{StaticResource MusicVisualizerMirroredGlyph}" />
                     </VisualState.Setters>
                 </VisualState>


Screenbox.Core/ViewModels/PlayerPageViewModel.cs
United600, line 575:
```suggestion
            if (focused is Slider) return false;
```

It seems like the `IsFocusEngaged` isn't doing what it's supposed to.

Screenbox/Pages/PlayerPage.xaml.cs
```diff
@@ -164,6 +163,19 @@ private void CoreTitleBar_LayoutMetricsChanged(CoreApplicationViewTitleBar sende
...
+        private void AccessKeyManager_OnIsDisplayModeEnabledChanged(object sender, object args)
+        {
+            ViewModel.KeyTipsVisible = AccessKeyManager.IsDisplayModeEnabled;
+            if (AccessKeyManager.IsDisplayModeEnabled)
+            {
+                ViewModel.ControlsHidden = false;
+            }
+            else
+            {
+                ViewModel.TryHideControls();
```
United600: A better approach is to use `DelayHideControls()`, though we'd have to expose it as public.

huynhsontung: We already have a pattern for handling this kind of thing in the view model. You can create a partial method in the view model to do view model related logic on property changed events.

```cs
partial void OnKeyTipsVisibleChanged(bool value)
{
    if (value)
        ControlsHidden = false;
    else
        TryHideControls();
}
```

Screenbox.Core/ViewModels/PlayerPageViewModel.cs
huynhsontung: What's the problem here? The focused engaged state is for gamepad interaction only.

United600: Oh! We should either add keyboard interaction support, or just use the the `DelayHideControls` to avoid immediate dismissal of the controls.

huynhsontung (approved):
LGTM

Screenbox/Pages/PlayerPage.xaml.cs
```diff
@@ -65,6 +58,8 @@ public PlayerPage()
 
             INavigationService navigationService = Ioc.Default.GetRequiredService<INavigationService>();   // For navigation events
             navigationService.Navigated += NavigationServiceOnNavigated;
+
+            AccessKeyManager.IsDisplayModeEnabledChanged += AccessKeyManager_OnIsDisplayModeEnabledChanged;
```
United600, line 62: What is your stance on using lambda expressions, especially for single-line cases?
I know that we can't unsubscribe from events when using them, but for now that doesn't look like a concern on this page, but should it be?

```suggestion
            AccessKeyManager.IsDisplayModeEnabledChanged += (sender, args) =>
            {
                ViewModel.KeyTipsVisible = AccessKeyManager.IsDisplayModeEnabled;
            };
```

Or even more compact:

```csharp
AccessKeyManager.IsDisplayModeEnabledChanged += (s, e) => { ViewModel.KeyTipsVisible = AccessKeyManager.IsDisplayModeEnabled; };
```

huynhsontung: It's better to have a named function because it helps with debugging once the application is deployed. If there is an exception, I can see the function name in the stack trace instead of just `Anonymous...()`.

PR #712, opened 2025-11-02
refactor: play queue selection logic and checkbox tooltip

United600:
- Adds a new observable property `SelectionCheckState` to `PlayQueueViewModel` to track the selection state for the play queue checkbox
- Removes `GetSelectionCheckBoxToolTip()`
- Refactors the selection tooltip by centralizing the logic on `CheckBoxToToolTipConverter` to workaround microsoft/microsoft-ui-xaml#1904
- Updates some `SelectionCheckBox` bindings to improve maintainability

### DIFF
--- a/Screenbox.Core/ViewModels/PlayQueueViewModel.cs
+++ b/Screenbox.Core/ViewModels/PlayQueueViewModel.cs
@@ -50,7 +50,7 @@ public sealed partial class PlayQueueViewModel : ObservableRecipient
         Playlist = playlist;
         _filesService = filesService;
         _resourceService = resourceService;
-        SelectionCheckState = false;
+        SelectionCheckState = GetSelectionCheckState(_selectionCount);
         _hasItems = playlist.Items.Count > 0;
         _dispatcherQueue = DispatcherQueue.GetForCurrentThread();
         Playlist.Items.CollectionChanged += ItemsOnCollectionChanged;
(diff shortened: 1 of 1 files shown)


Screenbox.Core/ViewModels/PlayQueueViewModel.cs
```diff
@@ -49,6 +50,7 @@ public PlayQueueViewModel(MediaListViewModel playlist, IFilesService filesServic
         Playlist = playlist;
         _filesService = filesService;
         _resourceService = resourceService;
+        SelectionCheckState = false;
```
United600: Do you think we should create a method to set the initial value instead of "cheating" with the _false_ value?

```csharp
private bool? GetSelectionCheckState(int selectionCount)
{
    return selectionCount == 0
        ? false
        : selectionCount == Playlist?.Items.Count ? true : null;
}
```

Then it can be reused in the `OnSelectionCountChanged()` method.

huynhsontung: Yea that makes sense

Screenbox/Converters/CheckBoxToToolTipConverter.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// <param name="value">The nullable boolean representing the checkbox state.</param>
+    /// <returns>
+    /// <b>SelectNoneToolTip</b> if <paramref name="value"/> is <see langword="true"/>;
+    /// <b>SelectAllToolTip</b> if <paramref name="value"/> is <see langword="false"/> or <see langword="null"/>.
+    /// </returns>
+    public static string GetSelectionToolTip(bool? value)
+    {
+        return value is null
+            ? Strings.Resources.SelectAllToolTip
+            : (value.Value ? Strings.Resources.SelectNoneToolTip : Strings.Resources.SelectAllToolTip);
```
United600, line 29: I was also thinking about using a switch expression to improve readability.

```suggestion
        return value switch
        {
            true => Strings.Resources.SelectNoneToolTip,
            false => Strings.Resources.SelectAllToolTip,
            null => Strings.Resources.SelectAllToolTip,
        };
```

huynhsontung: You could just do
 
```cs
return value is true ? Strings.Resources.SelectNoneToolTip : Strings.Resources.SelectAllToolTip;
```

United600: Not sure. In my eyes it's not entirely clear what **null** returns, and we may want to return a different string for **null** in the future.

huynhsontung: The value is a nullable Boolean. I would argue that the possible values are pretty clear. If there is a situation where we need to return a different string for the null case, we can change it in the future. Let's write our code to best fit the current requirements, not the future.

United600: You're right, this one is unlikely to change.

> Let's write our code to best fit the current requirements, not the future.

Unfortunately, reusability and maintainability is always top of my mind 😩.

PR #715, opened 2025-11-05
fix: UI not interactable when launching from explorer

huynhsontung:
`LightDismissLayer` in `NavigationViewEx` is visible in this path, blocking all UI interactions.

https://github.com/user-attachments/assets/e88f597f-87af-4bad-9a6d-01fa96a5b839

### DIFF
--- a/Screenbox/Controls/NavigationViewEx.cs
+++ b/Screenbox/Controls/NavigationViewEx.cs
@@ -254,10 +254,7 @@ public sealed partial class NavigationViewEx : NavigationView
                 }
 
                 var autoSuggestBox = AutoSuggestBox;
-                if (autoSuggestBox != null)
-                {
-                    autoSuggestBox.Focus(FocusState.Programmatic);
-                }
+                autoSuggestBox?.Focus(FocusState.Programmatic);
 
                 e.Handled = true;
                 return;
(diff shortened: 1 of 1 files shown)


United600 (approved):

Screenbox/Controls/NavigationViewEx.cs
United600, line 457:
```diff
- bool isLayerVisible = DisplayMode != NavigationViewDisplayMode.Expanded && IsPaneOpen;
-
- _overlayChildRectangle.Visibility = isLayerVisible ? Visibility.Visible : Visibility.Collapsed;
+ bool showLightDismissLayer =
+     (DisplayMode != NavigationViewDisplayMode.Expanded) &&
+     IsPaneOpen &&
+     (ContentVisibility == Visibility.Collapsed);
+ 
+ _overlayChildRectangle.Visibility = showLightDismissLayer ? Visibility.Visible : Visibility.Collapsed;
```

Adding the check here is probably clearer. No need for an extra method call.

huynhsontung: Can you explain this logic in plain English? When does the light dismiss layer show up?

United600: It should only be visible if the navigation view is in minimal/compact mode with the pane open. But it was missing the content visibility check that would cause this issue, and if we add it here we can remove this: https://github.com/huynhsontung/Screenbox/blob/43986e1bdd77d0b775d765be32de4f34b8162fda/Screenbox/Controls/NavigationViewEx.cs#L403

United600: I was totally wrong, sorry about that. I still think the `(ContentVisibility == Visibility.Visible)` (which is wrong on the example above, oops) check should be added.

PR #717, opened 2025-11-07
refactor: chapter control progress bar bindings

United600:


Screenbox/Controls/ChapterProgressBar.xaml
```diff
@@ -18,11 +18,11 @@
 
             <DataTemplate x:Key="ItemTemplate" x:DataType="viewModels:ChapterViewModel">
```
United600: Instead of spamming `Mode=OneWay` I could have just done:

```xml
<DataTemplate
    x:Key="ItemTemplate"
    x:DataType="viewModels:ChapterViewModel"
    x:DefaultBindMode="OneWay">
```

huynhsontung: Ooh I don't know about that. I think explicitly state binding mode still has its merits. I rely on the fact that x:Bind defaults to OneTime quite often. I don't know if implicitly changing the binding mode is a good idea here.

United600: Yeah, it still seems useful in certain situations (like in this one).

huynhsontung (commented):
If I remember correctly, I used Bindings here for a reason. When I last tested this, there were some edge cases where ItemRepeater and x:Bind don't play nice together. Did you encounter any issues in your testing?

United600:
> If I remember correctly, I used Bindings here for a reason. When I last tested this, there were some edge cases where ItemRepeater and x:Bind don't play nice together. Did you encounter any issues in your testing?

No, but I haven't tested it extensively. Just a couple of files with chapters. But now that I know you've run into issues, I'll test it with a few more.

United600:
> If I remember correctly, I used Bindings here for a reason. When I last tested this, there were some edge cases where ItemRepeater and x:Bind don't play nice together. Did you encounter any issues in your testing?

Can confirm. Works correctly for the first file in a session, but subsequent files incorrectly applies the `Value` to some chapters.

The `ItemsRepeater` also isn't properly clearing the list. While replacing it with an `ItemsControl` would resolve both this issue and the previous one, it introduces a regression (increasing the `Spacing` will cause value misalignment in some files).

```xml
<ItemsControl ItemTemplate="{StaticResource ItemTemplate}" ItemsSource="{x:Bind ProgressItems}">
    <ItemsControl.ItemsPanel>
        <ItemsPanelTemplate>
            <StackPanel Orientation="Horizontal" Spacing="1" />
        </ItemsPanelTemplate>
    </ItemsControl.ItemsPanel>
</ItemsControl>
```

huynhsontung:
I have no concern for changing from `ItemsRepeater` to `ItemsControl`.

United600:
> I have no concern for changing from `ItemsRepeater` to `ItemsControl`.

Same here, I just need to test it a bit more.

Screenbox/Controls/ChapterProgressBar.xaml.cs
```diff
@@ -268,7 +269,7 @@ private void PopulateProgressItems()
 
         private double GetItemWidth(double durationMs, int chapterCount)
         {
-            double availableWidth = ActualWidth - Spacing * chapterCount;
+            double availableWidth = ActualWidth - _chapterSpacing * chapterCount;
```
United600:
```suggestion
            double availableWidth = ActualWidth - _chapterSpacing * (chapterCount - 1);
```

This should be `(chapterCount - 1)`, but I've learned my lesson trying to fix it. The whole setup is far too fragile, and floating‑point imprecision leads to incorrect calculations and occasionally cuts off the end (my assumption).

huynhsontung: No, this is intentional. I remember being burned by this. Without the extra space, the alignment was off. I think there was some padding inside the Slider control and the extra spacing compensated that padding. Not sure how that would work now that character spacing is a variable.

United600: I was burned by it too. If we really need extra spacing, it should be applied as its own explicit value, at least for formula correctness.
The Slider control has zero padding 🤔, so my theory was that floating‑point accumulation was responsible for the last item getting cut off

> Not sure how that would work now that character spacing is a variable.

Yeah, my bad for turning that constant into a resource, it was the only workaround I knew to keep both values synchronized, since `ItemsPanelTemplate` doesn't support `x:Bind`.

huynhsontung: Maybe we can just adjust it so the constant 1 px is retained. Delay the troubleshooting once again 😃 

```cs
const int AlignmentAdjustment = 1;
availableWidth = ActualWidth - _chapterSpacing * (chapterCount - 1) - AlignmentAdjustment;
```

United600: I guess, at least now anyone looking at it can immediately grasp the intent and won't waste time heading off in the wrong direction.

huynhsontung: My bad on that 😅

PR #719, opened 2025-11-09
feat: clear search box with Escape key

United600:
- Adds a new keyboard shortcut (Escape key) to clear the search box and its suggestions
- Renames the focus keyboard accelerator event handler from `NavViewSearchBoxKeyboardAcceleratorFocus_OnInvoked` to `NavViewSearchBoxFocusKeyboardAccelerator_OnInvoked` for consistency (Control+Action+Event)

Not perfect, as it often eats the first press of the Escape key.

### DIFF
--- a/Screenbox/Pages/MainPage.xaml
+++ b/Screenbox/Pages/MainPage.xaml
@@ -230,8 +230,12 @@
                     <AutoSuggestBox.KeyboardAccelerators>
                         <KeyboardAccelerator
                             Key="F"
-                            Invoked="NavViewSearchBoxKeyboardAcceleratorFocus_OnInvoked"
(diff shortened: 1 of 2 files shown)


huynhsontung:
I don't think pressing Escape to clear the search box is an expected Windows behavior?

United600:
> I don't think pressing Escape to clear the search box is an expected Windows behavior?

Isn't it? I think settings search box, or start menu, clear with escape. Or was it the File Explorer/browser?

huynhsontung:
No, Settings and inbox Media Player definitely don't do that. File Explorer does though.

United600:
Right, the Store also doesn't. It might just be because the default `TextBox` doesn't handle it.

huynhsontung:
Closing this since the expected behavior is unclear. We should stick to WinUI behavior.

PR #733, opened 2025-11-14
refactor: player page keyboard accelerator events

United600:
Refactors keyboard accelerator handling `On*KeyboardAcceleratorInvoked` methods in `PlayerPageViewModel` and updates `PlayerPage.xaml` and `PlayerPage.xaml.cs` to use new handlers for volume control, seeking, percent jumps, playback rate, frame stepping, and window resizing.


huynhsontung:
Hmm, don't you still need to define the `KeyboardAccelerators` in XAML in addition to handling the `ProcessKeyboardAccelerators` event?

United600:
> Hmm, don't you still need to define the `KeyboardAccelerators` in XAML in addition to handling the `ProcessKeyboardAccelerators` event?

Surprisingly, no.

United600:
But if it's too much "magic", I can revert [8a170a1](https://github.com/huynhsontung/Screenbox/pull/733/commits/8a170a1de0ea86e6e8d705e8d4cdb4cb26f04240)

United600:
You might be right. After testing it a bit more, I can see that it isn't reliable whenever the player isn't in focus.

huynhsontung:
> You might be right. After testing it a bit more, I can see that it isn't reliable whenever the player isn't in focus.

That's...not the behavior I expect it to have lol. Oh well, more XAML mystery.

United600:
> That's...not the behavior I expect it to have lol. Oh well, more XAML mystery.

If any of the player buttons is focused it worked flawlessly, but as soon as it moved to another control it would stop processing them. It's a shame because it was really clean but the minimal player visual state didn't cooperate.

Screenbox.Core/ViewModels/PlayerPageViewModel.cs
```diff
@@ -430,30 +510,17 @@ public void OnResizeKeyboardAcceleratorInvoked(KeyboardAccelerator sender, Keybo
...
-            case (VirtualKey)0xBB when sender.Modifiers == VirtualKeyModifiers.Control:  // Plus ("+")
-                ResizeWindow(currentSize, 1 + desiredStepSize);
-                break;
-            case (VirtualKey)0xBD when sender.Modifiers == VirtualKeyModifiers.Control:  // Minus ("-")
-                ResizeWindow(currentSize, 1 - desiredStepSize);
-                break;
-            default:
-                args.Handled = false;
-                break;
-        }
```
huynhsontung, line 477: Very nice. I like this change!

Screenbox/Pages/PlayerPage.xaml.cs
```diff
@@ -109,7 +109,6 @@ protected override void OnKeyDown(KeyRoutedEventArgs e)
                     VideoView.ContextFlyout.ShowAt(PlayerControls,
                         new FlyoutShowOptions { Placement = GlobalizationHelper.IsRightToLeftLanguage ? FlyoutPlacementMode.TopEdgeAlignedLeft : FlyoutPlacementMode.TopEdgeAlignedRight });
                     break;
-                case VirtualKey.Escape when shouldHideControls:
```
huynhsontung, line 112: Maybe I missed some context. Why do we want to remove this condition?

huynhsontung (commented):
Very nice. I like the direction we are going with separating view components from view mode logic.

Screenbox/Pages/PlayerPage.xaml.cs
United600: https://github.com/huynhsontung/Screenbox/blob/94487281efc7eadbbce0a76589124e78a6fc1545/Screenbox/Pages/PlayerPage.xaml#L168

I uncommented that bit and added the following:

https://github.com/huynhsontung/Screenbox/pull/733/files#diff-484836b95b66669936358e6d50d0df20b48d4727688b6e02a224fe02a3e20851R545-R552

TBF, I might be missing some context on why it was set up that way initially. Probably a good candidate for the `ProcessKeyboardAccelerators` event since it requires focus to work.

Screenbox.Core/ViewModels/PlayerPageViewModel.cs
United600: Fundamentally it didn’t actually change anything, but thanks anyway. Yay for better syntax.

Screenbox/Pages/PlayerPage.xaml.cs
huynhsontung: It's been so long, idk what happened here. It may have to do with the Escape key not behaving as expected in a KeyboardAccelerator context. Can you help me verify on Xbox? If this behaves as expected on Xbox then we can push this.

United600: Doesn't really touch any Xbox specific behavior. Plus I would need to get the Xbox Dev Mode, which isn't a major hurdle on its own, but I'm more concerned in figuring out how to deploy and debug it.

huynhsontung: Oh I though you always had Xbox Dev mode. Don't worry about this then.

PR #748, opened 2025-12-03
doc: Add Stargazers section to README

huynhsontung:
A fun addition 🌟


United600:
Meh. Actually it's a bit bland and doesn't look too good.

huynhsontung:
Too bad

United600:
Not opposed to it, but it doesn't really provide any value to the `README`. If at least it had some customization to not feel so disjointed.

PR #750, opened 2025-12-07
ci: Change release type from 'published' to 'released'

drshtmstry:


huynhsontung (approved):
Confirm with [GHA events doc](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#release). Thank you for the fix.

PR #757, opened 2025-12-13
feat: add holding and swipe player gestures

### LINKED ISSUE #866: hold pause for 2x playback
it'd be nice if there was a feature to hold spacebar or lmb to temporarily switch to 2x playback or the desired speed in settings

United600:
Refactors the gesture handling system for the media player, making it more flexible and configurable. Instead of using simple boolean flags for gestures like volume and seek, it now allows mapping various gestures (tap and swipe directions) to specific media commands via a new `PlaybackActionKind` enum. This improves extensibility and allows for more customized user interactions.

Adds press and hold gesture support (boosts playback speed)

It also refactors the `VideoView.PointerWheelChanged` and `VlcVideoView.SizeChanged` methods to the view, and adds support for horizontal touchpad/mouse wheel input.

Gesture settings

Gesture settings with the options expanded


Screenbox.Core/ViewModels/SettingsPageViewModel.cs
```diff
@@ -58,6 +63,8 @@ public sealed partial class SettingsPageViewModel : ObservableRecipient
 
     public int[] PlayerControlsHideDelayOptions { get; } = { 1, 2, 3, 4, 5 };
 
+    public Array GestureOptions { get; }
```
huynhsontung: This is a weird type. You meant `object[]`?

United600: No, I just copied the return type from the Community Toolkit [EnumValuesExtensions](https://github.com/CommunityToolkit/Windows/blob/main/components/Extensions/src/Markup/EnumValuesExtension.cs). If you prefer we could've use the extension to set the `Enum` directly as the `ItemSource`.

Screenbox/Controls/PlayerElement.xaml.cs
```diff
@@ -53,4 +85,69 @@ private void VideoViewButton_OnDoubleTapped(object sender, DoubleTappedRoutedEve
...
+    private void VideoViewButton_OnPointerCanceled(object sender, PointerRoutedEventArgs e)
+    {
+        _gestureRecognizer.CompleteGesture();
+        VideoViewButton.ReleasePointerCapture(e.Pointer);
+        e.Handled = IsEnabled;
+    }
+
+    private void VideoViewButton_OnPointerMoved(object sender, PointerRoutedEventArgs e)
+    {
+        if (!IsEnabled) return;
```
United600:
```suggestion
        if (!IsEnabled || ViewModel.IsHolding) return;
```

Screenbox.Core/ViewModels/PlayerPageViewModel.cs
```diff
@@ -290,9 +293,66 @@ public void OnPointerMoved()
...
+            // If the rate is already faster than the holding speed, set it to twice the holding speed.
+            double effectiveHoldingSpeed = MediaPlayer.PlaybackRate >= HoldingSpeed ? HoldingSpeed * 2.0 : HoldingSpeed;
+            if (MediaPlayer.PlaybackRate != effectiveHoldingSpeed)
+            {
+                Messenger.Send(new ChangePlaybackRateRequestMessage(effectiveHoldingSpeed));
+                Messenger.Send(new UpdateStatusMessage(Humanizer.FormatPlaybackRate(effectiveHoldingSpeed), System.Threading.Timeout.InfiniteTimeSpan));
+            }
+
+            _isSpaceKeyHolding = true;
+        }, TimeSpan.FromMilliseconds(300));
```
United600: I couldn't find official documentation specifying the default keyboard repeat delay, but 500 ms seemed consistent with the behavior of the other inputs (touch and mouse).

Screenbox.Core/Services/SettingsService.cs
```diff
@@ -38,6 +35,13 @@ public sealed class SettingsService : ISettingsService
     private const string PlayerShowChaptersKey = "Player/ShowChapters";
     private const string PrivacyPersistPlaybackPosition = "Privacy/PersistPlaybackPosition";
 
+    private const string PlayerGestureTapKey = "Player/Gesture/Tap";
```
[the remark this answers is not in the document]
huynhsontung: Casting `bool` to `int` is planned. True value will be cast to 1, which matches `PlaybackActionKind.PlayPause`. Maintain backward compatibility.

Screenbox.Core/Enums/PlaybackActionKind.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace Screenbox.Core.Enums;
+
+/// <summary>
+/// Represents the action that was taken on the media playback.
+/// </summary>
+public enum PlaybackActionKind
```
United600, line 6: Should we consider converting this enum to a flag enum (and renaming it to `PlaybackActionKinds`) now? This would allow us to add pointer wheel (and slide gesture) customization, though we'd need to update the `ResourceNameToResourceStringConverter` but that shouldn't be too difficult.

Then we could do something like this:

```csharp
// SettingsPageViewModel.cs
public PlaybackActionKind[] PointerWheelOptions { get; }

...

var pointerWheelSeekAction = PlaybackActionKind.Rewind | PlaybackActionKind.FastForward;
var pointerWheelVolumeAction = PlaybackActionKind.DecreaseVolume | PlaybackActionKind.IncreaseVolume;

PointerWheelOptions = new[]
{
    PlaybackActionKind.None,
    pointerWheelSeekAction,
    pointerWheelVolumeAction,
};

// PlayerElementViewModel.cs
private void ProcessPlayerGesture(PlaybackActionKind gestureOption, double change)
{
    ...

    switch (gestureOption)
    {
        ...

        case PlaybackActionKind.Rewind | PlaybackActionKind.FastForward:
            ...
            break;
        case PlaybackActionKind.DecreaseVolume | PlaybackActionKind.IncreaseVolume:
            ...
            break;
    }
}
```

huynhsontung: I'm not sure I see the benefits of making this enum a flag. Especially when it's user facing since it involves resource string. Should the slide gesture use the same enum? Maybe it warants its own enum since the behavior of slide gesture is fundamentally different from other gestures. Even if we were to use the same enum then can't we add dedicated entries to the enum instead of using calculated ones?

United600: > Should the slide gesture use the same enum? Maybe it warants its own enum since the behavior of slide gesture is fundamentally different from other gestures. Even if we were to use the same enum then can't we add dedicated entries to the enum instead of using calculated ones?

Are they really that different? My thinking is along the lines of keeping all the gesture-related logic within the `ProcessPlayerGesture` method, but once we ship this there's no turning back with that enum.

> I'm not sure I see the benefits of making this enum a flag. Especially when it's user facing since it involves resource string.

I don't think this will be a major problem. We can check for combined values, allowing us to use a resource string like this:

```xml
<data name="PlaybackActionKindDecreaseVolumeIncreaseVolume" xml:space="preserve">
  <value>Change volume</value>
</data>
```

United600: > Even if we were to use the same enum then can't we add dedicated entries to the enum instead of using calculated ones?

Sure, though it wouldn't be as "elegant" and would require updating the `GestureOptions` in the `SettingsViewModel` (which isn't a major issue).

huynhsontung: I still don't think we should make it a flag. I still hold to my recommendation that we either create a new enum or add dedicated values to the existing `PlaybackActionKind`.

United600: I guess it's fine, and we avoid some complexity when dealing with the converter. Just wanted to make sure we don't regret it later.

huynhsontung:
If there is no other concern, I will merge this PR. It's getting too big that merging other PRs may cause merge conflicts with this one.

Screenbox/Strings/en-US/Resources.resw
```diff
@@ -664,13 +664,10 @@
...
   <data name="SettingsGestureSeek" xml:space="preserve">
-    <value>Drag horizontally to seek</value>
+    <value>Enable slide horizontally to seek</value>
   </data>
   <data name="SettingsGestureVolume" xml:space="preserve">
-    <value>Drag vertically to adjust volume</value>
-  </data>
-  <data name="SettingsGestureTap" xml:space="preserve">
-    <value>Tap anywhere to play or pause</value>
+    <value>Enable slide vertically to adjust volume</value>
```
huynhsontung: These changes our language a little bit. Adding "Enable" at the start sounds too programmatic. Maybe because we have two verbs next to each other? "Enable slide...". I prefer not having "Enable..."

United600: I took inspiration from this system setting, but it's okay to remove it. Do you think we also should change the press and hold string to something like "Press and hold temporarily increases the playback speed", or "Press and hold to temporarily increase the playback speed"? The second option seems off without "Enable...".

Windows 11 additional touch settings

United600:
> If there is no other concern, I will merge this PR. It's getting too big that merging other PRs may cause merge conflicts with this one.

Agreed, it's ready. A couple of minor UI tweaks might be needed, but can be addressed later.

Screenbox.Core/ViewModels/PlayerElementViewModel.cs
```diff
@@ -30,6 +29,17 @@ public sealed partial class PlayerElementViewModel : ObservableRecipient,
     IRecipient<ChangeAspectRatioMessage>,
     IRecipient<SettingsChangedMessage>
 {
+    private const double GestureStepAmount = 5.0;
+
+    private enum SlideDirection
```
United600:
```suggestion
    private enum SlideAxis
```

To me, direction means left, right, up, down. `SlideRail`? `SlideOrientation`? I think "Axis" and "Rail" lean more toward "X/Y/Z", so "Orientation" might be the better choice.

Screenbox/Strings/en-US/Resources.resw
huynhsontung: "Press and hold to temporarily increase the playback speed" sounds off to me, too. "Press and hold temporarily increases the playback speed" is also not technically correct English, either. Maybe for press and hold, we follow the system settings and add "Enable..." at the front. Keep the slide strings without.

Screenbox.Core/ViewModels/PlayerElementViewModel.cs
huynhsontung: I agree. `SlideOrientation` is good for me. Good catch.

Screenbox/Strings/en-US/Resources.resw
United600: So, "**Enable press and hold to increase playback speed**", or "**Enable press and hold to temporarily increase playback speed**"? Does "press and hold" clearly indicate that the action is temporary, or should we include "temporarily"? I think we can omit it.

Other alternative, "Enable press and hold to speed up playback".

Screenbox.Core/ViewModels/PlayerElementViewModel.cs
United600: 🤔 We could make `_slideOrientation` nullable and use the platform [Orientation](https://learn.microsoft.com/uwp/api/windows.ui.xaml.controls.orientation) enum. Maybe not, but I should've used that instead of a nullable bool.

Screenbox/Strings/en-US/Resources.resw
huynhsontung: I think we can skip "temporarily". So "Enable press and hold to increase playback speed" is good to me 👍

Screenbox.Core/ViewModels/PlayerElementViewModel.cs
huynhsontung: No, that's fine. We are not saving much by reusing an enum :)

PR #769, opened 2026-01-07
fix: RTL repeat glyph directionality

United600:
Removes the RepeatAll and RepeatOff mirrored glyphs and fixes the RepeatOneMirrored glyph.

|Before | After|
|-|-|
|Before|After|

Thanks for the advice, @Avid29.


Avid29:
For the record, this is not my advice. It's a modifier, not a badge. Just use the same icon.

United600:
> For the record, this is not my advice. It's a modifier, not a badge. Just use the same icon.

Right, it was about the glyph direction, specifically, it should be clockwise.

PR #775, opened 2026-01-19
refactor: add playback rate message

United600:
Adds a new message-based approach for playback rate updates, unifies the terminology from "speed" to "rate", and expands the available playback rate options (4x).

Also adds constants for OEM keys (comma, period, plus, minus).


Screenbox.Core/Messages/ChangePlaybackRateRequestMessage.cs
```diff
@@ -0,0 +1,23 @@
+﻿using CommunityToolkit.Mvvm.Messaging.Messages;
+
+namespace Screenbox.Core.Messages;
+
+/// <summary>
+/// Represents a request message to change the playback rate of the media player.
+/// </summary>
+public sealed class ChangePlaybackRateRequestMessage : RequestMessage<double>
+{
+    /// <summary>
```
United600, line 10:
```suggestion
{
    public const double MinRate = 0.05;
    public const double MaxRate = 4.0;
    public const double Step = 0.25;
    
    /// <summary>
```

Is it a good idea to move the constants to the message?

Screenbox.Core/ViewModels/PlayerControlsViewModel.cs
United600:
```suggestion
                    AudioTimingOffset = 0;
                    PlaybackRate = 1.0;
```

I wonder if we should reset the playback rate on media change? I don't think so, but maybe behind a setting? Probably not even that.

Screenbox.Core/Messages/ChangePlaybackRateRequestMessage.cs
huynhsontung: I don't think so. These are more of the implementation details. I think they should either be in the view model or a separate static class. Or just not expose these values. I don't think the controls need these values anyway.

United600: Are you suggesting that the clamping on the message `Receive` isn't the best approach? I can see that, the controls already handle that. I'd need to apply the clamp on the `Shift` + `.`/`,` shortcut.

United600: Ah! Clamping it would still be useful, so we can set the minimum on the slider to 0x, making 1x the midpoint.

huynhsontung: Should the minimum even be 0x? Will it break anything if the playback rate is 0x? I think we should clamp it either at the control level or at the view model level. I don't have a preference where.

Screenbox.Core/ViewModels/PlayerControlsViewModel.cs
huynhsontung: Oh yeah, we currently don't reset the playback rate on media change. My initial instinct is we should reset it after every media change, but I can't really explain why. The current behavior makes sense too logically. I think we should leave it as-is. This value only persists during the app session anyway.

Screenbox.Core/Messages/ChangePlaybackRateRequestMessage.cs
United600: I'm not sure, but the goal is to set the `Minimum="0"` on the slider to fix this current visual imbalance and let the ViewModel/Message handle the clamping.

Slider with tick frequency of 0.25

huynhsontung: I get that. Maybe we should just put a number box there and let people enter the rate they want.

United600: IIRC using NumberBox with a gamepad was horrible, if not outright broken.

PR #784, opened 2026-02-21
style: add titlebar control

United600:
Replaces custom title bars with reusable TitleBar control.

- Adds a new modular `TitleBar` control supporting header, content, footer, icon, and title properties, as well as customizable caption button colors and height modes
- Removes legacy title bar code and updates `MainPage` and `PlayerPage` to use the new control
- Replaces `MainPage` and `PlayerPage` drag region and caption button color updates with `TitleBar` control
- Removes theme and caption button color handling logic from `App` initialization
- Adds `NavigationView` CompactPaneOverlay visual state
- Removes `TitleBarHelper` and related code
- Makes the media info in the MiniPlayer clickable (returns to the immersive view)


Screenbox/Controls/TitleBar/TitleBar.cs
```diff
@@ -0,0 +1,414 @@
...
+    private Window? _window;
+
+    private long _flowDirectionCallbackToken;
+
+    private ColumnDefinition? _leftPaddingColumn;
+    private ColumnDefinition? _rightPaddingColumn;
+    private Grid? _dragRegion;
+    private FrameworkElement? _headerArea;
+    private FrameworkElement? _contentArea;
+    private FrameworkElement? _footerArea;
```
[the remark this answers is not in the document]
United600: They are being used.

United600:
Should we add the icon during compact overlay mode?

Screenbox playing music in compact mode

Screenbox/Pages/MainPage.xaml.cs
```diff
@@ -318,7 +295,10 @@ private void NavView_OnPaneOpening(NavigationView sender, object args)
 
         private void NavView_OnPaneClosing(NavigationView sender, object args)
         {
-            UpdateNavigationViewState(sender.DisplayMode, sender.IsPaneOpen);
+            // Deferred to ensure IsPaneOpen reports the correct state
+            // when closing the pane via gamepad.
+            _ = Dispatcher.RunAsync(CoreDispatcherPriority.Normal,
+                () => UpdateNavigationViewState(sender.DisplayMode, sender.IsPaneOpen));
```
United600, line 302: Maybe be a bit excessive, but I wanted to avoid adding another event handler (PaneClosed).

What if we just pass the value directly? That way, we can skip using the `Dispatcher`?
```diff
- UpdateNavigationViewState(sender.DisplayMode, sender.IsPaneOpen);
+ UpdateNavigationViewState(sender.DisplayMode, false);
```

huynhsontung:
> Should we add the icon during compact overlay mode?
 
I think so. Just checked YouTube and the inbox media player, and they do show their icon.

Screenbox/Pages/MainPage.xaml.cs
huynhsontung: Did you try this? I think some NavigationView visual state changes must happen first before we can update the state on our end. If this doesn't produce unexpected visual behavior then I am okay.

United600: I tried it, and it worked well, but it could use a bit more testing.

Screenbox/Themes/Generic.xaml
huynhsontung, line 1: Oh we are using `Generic.xaml` now. I know this will be loaded on app launch, but it still feels very implicit. It's not immediately clear when this file is loaded since it's not being referenced anywhere in other XAML files. Why can't we load this in `App.xaml`?

United600: Not sure. Templated controls generate that file, and it's standard practice from what I gathered. Should we attempt to remove it? Is that even possible?

huynhsontung: Just looked up. `Generic.xaml` allows dependent projects to consume as the default style for the controls defined there. It makes sense if you are building a control library. If the project is standalone (such as our case), then the result is the same. Hmm I'm okay with this. The question now is, do we move other templated controls to `Generic.xaml` or should we stick with `App.xaml`?

United600: Probably best to stick with the `Generic.xaml` in case we migrate the controls to a dependent project later. I also intend to create/convert more controls into templated controls.

> The question now is, do we move other templated controls to Generic.xaml or should we stick with App.xaml?

Do you mean the `NavigationViewEx` and `CommandBarEx`? For now `TitleBar` is the only "full" templated control.

PR #801, opened 2026-03-13
refactor: language model and remove layout direction converter

United600:
- Updates language selection `ComboBox` to use `TextAlignment` instead of `FlowDirection`
- Removes `LanguageLayoutDirectionToFlowDirectionConverter` and related logic, as it's no longer necessary
- Removes the `LayoutDirection` property from the language model
- Renames `Language` to `LanguageInfo` to resolve ambiguity


Screenbox/Pages/SettingsPage.xaml
```diff
@@ -304,8 +302,8 @@
                         ItemsSource="{x:Bind ViewModel.AvailableLanguages}"
                         SelectedIndex="{x:Bind ViewModel.SelectedLanguage, Mode=TwoWay}">
                         <ComboBox.ItemTemplate>
-                            <DataTemplate x:DataType="models:Language">
-                                <TextBlock FlowDirection="{x:Bind converters:LanguageLayoutDirectionToFlowDirectionConverter.ToFlowDirection(LayoutDirection)}" Text="{x:Bind NativeName}" />
+                            <DataTemplate x:DataType="models:LanguageInfo">
+                                <TextBlock Text="{x:Bind NativeName}" TextAlignment="DetectFromContent" />
```
huynhsontung, line 306: Did not know TextAlignment has this `DetectFromContent`... Could have saved the trouble from the beginning 😅

United600: Ahah. I also think that we should use it as the default in the `CommonGridViewItem.HorizontalTextAlignment`.

PR #804, opened 2026-03-14
style: improve playlist button design in the play queue selection bar

United600:
Changes `AddToPlaylistSelectionButton` to define flyout chevron visibility, glyph, font size, and margin for improved flyout appearance.

**Normal state**
Selection command bar featuring the 'Add to Playlist' option with the added chevron

**Overflow state**
Selection command bar featuring the 'Add to Playlist' option in the overflow menu, and with the added chevron

Also fixes a potential overlap in the `PlayQueuePage` header

https://github.com/user-attachments/assets/ec1d2a22-b27d-473e-b788-c71efb183f96

### DIFF
--- a/Screenbox/Pages/PlayQueuePage.xaml
+++ b/Screenbox/Pages/PlayQueuePage.xaml
@@ -59,14 +59,26 @@
             Grid.Row="0"
             MinHeight="{StaticResource PageHeaderMinHeight}"
             Margin="{StaticResource BottomMediumMargin}"
-            Padding="{StaticResource ContentPagePadding}">
+            Padding="{StaticResource ContentPagePadding}"
+            ColumnSpacing="8">
+            <Grid.ColumnDefinitions>
+                <ColumnDefinition Width="*" />
+                <ColumnDefinition Width="Auto" />
+            </Grid.ColumnDefinitions>
+
             <TextBlock
-                HorizontalAlignment="Left"
+                Grid.Column="0"
                 Style="{StaticResource TitleMediumTextBlockStyle}"
-                Text="{strings:Resources Key=PlayQueue}" />
+                Text="{strings:Resources Key=PlayQueue}"
+                TextWrapping="NoWrap">
+                <interactivity:Interaction.Behaviors>
+                    <behaviors:OverflowTextToolTipBehavior />
+                </interactivity:Interaction.Behaviors>
+            </TextBlock>
+
             <muxc:SplitButton
                 x:Name="HeaderAddButton"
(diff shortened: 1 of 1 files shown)


United600:
BTW I'm not a fan of the dropdown button in the actions row. We should be careful about potential button truncation in some languages, especially since it hasn't been converted to a command bar yet.

huynhsontung:
Why would there be truncation? If the label is too long, the commands will just be pushed into the secondary command dropdown, right?

Screenbox/Controls/PlayQueueControl.xaml
```diff
@@ -228,9 +228,15 @@
...
                     x:Name="AddToPlaylistSelectionButton"
-                    Icon="{ui:SymbolIcon Symbol=Add}"
+                    Icon="Add"
                     IsEnabled="{x:Bind ViewModel.SelectionCount, Converter={StaticResource DoubleToBoolConverter}, Mode=OneWay}"
                     Label="{strings:Resources Key=AddToPlaylist}">
+                    <AppBarButton.Resources>
+                        <Visibility x:Key="AppBarButtonHasFlyoutChevronVisibility">Visible</Visibility>
+                        <x:String x:Key="AppBarButtonFlyoutGlyph">&#xE972;</x:String>
+                        <x:Double x:Key="AppBarButtonSubItemChevronFontSize">12</x:Double>
+                        <Thickness x:Key="AppBarButtonSubItemChevronLabelOnRightMargin">-7,18,12,0</Thickness>
```
huynhsontung, line 238: Is it because there is a negative left margin here?

United600:
> Why would there be truncation? If the label is too long, the commands will just be pushed into the secondary command dropdown, right?

I'm talking about the button in the actions row not the selection command bar. Sorry for the confusion.

Screenbox/Controls/PlayQueueControl.xaml
United600: That margin was like that already, I've only reduced the top field (from 20) to center the icon since the size was increased.

huynhsontung:
> I'm talking about the button in the actions row not the selection command bar. Sorry for the confusion.

I see. I wonder if it's better to start converting all actions row to use CommandBar for this reason?

United600:
> I see. I wonder if it's better to start converting all actions row to use CommandBar for this reason?

That was my idea and now that we have the CommandBar resource dictionary we can replicate the buttons look (like we do in the details pages).

German is usually a good language to test, and for now seems fine (but still yet to translate).

Play queue page in the minimal state

Play queue page in the player state

PR #813, opened 2026-03-17
style: update player page to show window deactivation status

United600:
Added Deactivated visual state to `PlayerPage.xaml` so when the window is deactivated, the title bar text colors are updated, especially important for the compact overlay mode.
I wonder if we should consider lowering the opacity of the top buttons and controls?

https://github.com/user-attachments/assets/69b7fcaf-6db9-46ba-9af6-5c60e86eaffc

### DIFF
--- a/Screenbox/Triggers/WindowActivationModeTrigger.cs
+++ b/Screenbox/Triggers/WindowActivationModeTrigger.cs
@@ -1,95 +1,93 @@
 ﻿#nullable enable
 
+using CommunityToolkit.WinUI.Helpers;
 using Windows.UI.Core;
 using Windows.UI.Xaml;
 
 namespace Screenbox.Triggers;
 
 /// <summary>
-/// Represents a declarative rule that applies visual states based on the <see cref="CoreWindow.ActivationMode"/> property.
+/// Represents a declarative rule that applies visual states based on the
+/// <see cref="CoreWindow.ActivationMode"/> property.
 /// </summary>
 /// <remarks>
-/// Use WindowActivationModeTriggers to create rules that automatically triggers a VisualState change when
-/// the window is a specified activation state. When you use WindowActivationModeTriggers in your XAML markup,
-/// you don't need to handle the <see cref="CoreWindow.Activated"/> event and call <see cref="VisualStateManager.GoToState"/> in your code.
+/// Use WindowActivationModeTriggers to create rules that automatically triggers a
+/// <see cref="VisualState"/> change when the window is a specified activation state.
+/// When you use WindowActivationModeTriggers in your XAML markup, you don't need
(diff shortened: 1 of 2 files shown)


huynhsontung:
Yes, we should dim the top controls and buttons, similar to how the NavView dims the back button when the window is deactivated.

United600:
> Yes, we should dim the top controls and buttons, similar to how the NavView dims the back button when the window is deactivated.

I'm concerned about the hover visuals, because unlike the NavView back button they'll still react. 
We could disable the title bar buttons, but that introduces its own issues such as requiring the window to be focused before they respond to any input.

Here is the related [documentation](https://learn.microsoft.com/en-us/windows/uwp/ui-input/title-bar#dim-the-title-bar-when-the-window-is-inactive) and pertinent point.

Title bar do's and dont's

PS: I don't know of a single app that dims the icon.

Screenbox/Triggers/WindowActivationModeTrigger.cs
```diff
@@ -1,95 +1,93 @@
...
+    /// Initializes a new instance of the <see cref="WindowActivationModeTrigger"/> class.
+    /// </summary>
+    public WindowActivationModeTrigger()
     {
-        _coreWindow = Window.Current?.CoreWindow;
+        _coreWindow = Window.Current.CoreWindow;
 
-        var coreWindow = _coreWindow;
-        if (coreWindow != null)
+        _weakEventListener = new(this)
```
United600, line 80: Inspired by the Community Toolkit [NetworkConnectionStateTrigger](https://github.com/CommunityToolkit/Windows/blob/v8.2.251219/components/Triggers/src/NetworkConnectionStateTrigger.cs) trigger, and using [this helper](https://github.com/CommunityToolkit/Windows/blob/main/components/Helpers/src/WeakEventListener.cs).

Is this the best way to "unsubscribe" from the event? Not overly familiar with [weak events](https://www.meziantou.net/weak-events-in-csharp.htm).

huynhsontung: Yes. This is a perfect use of WeakEventListener, actually. This is great.

huynhsontung:
Technically, `MainPage` and `PlayerPage` will persist throughout the app session, so the Trigger's lifetime isn't that important, since it must live until the app terminates. But this is a good standard.

United600:
> Technically, `MainPage` and `PlayerPage` will persist throughout the app session, so the Trigger's lifetime isn't that important, since it must live until the app terminates. But this is a good standard.

True, thought I really want to try to unload the `PlayerPage` when not in use. Ideally both pages wouldn't be coupled together, but that ship has sailed with the way we handle the `Overlay`.

PR #816, opened 2026-03-19
fix: album and artist search uses the internal key lead to unexpected matches

United600:
The album search was indexing using a key that included the artist name, leading to crashes during the suggestion ranking.


huynhsontung:
This is indeed a regression from #761

PR #825, opened 2026-03-27
feat: enable .NET Native trimming

United600:
Followed [this blog post](https://devblogs.microsoft.com/ifdef-windows/leveraging-trimming-to-make-the-microsoft-store-faster-and-reduce-its-binary-size/), and used [this troubleshooter](https://dotnet.github.io/native/troubleshooter/type.html) to enable .NET Native trimming.

Reduces the overall bundle size by approximately 10 MB.
Comparison between a bundle with trimming and one without it

And after unzipping, each package saves approximately 10 MB, with the `Screenbox.dll` file being the primary beneficiary of this reduction.

| Without trimming | With trimming |
|-|-|
|imagem|imagem|

### DIFF
--- a/Screenbox/Properties/Default.rd.xml
+++ b/Screenbox/Properties/Default.rd.xml
@@ -21,11 +21,37 @@
       An Assembly element with Name="*Application*" applies to all assemblies in
       the application package. The asterisks are not wildcards.
     -->
-    <Assembly Name="*Application*" Dynamic="Required All" />
+    <!--<Assembly Name="*Application*" Dynamic="Required All" />-->
     
     
     <!-- Add your application specific runtime directives here. -->
+    
+    <Namespace Name="ProtoBuf.Serializers" Dynamic="Required Public" />
+    <!--<Type Name="ProtoBuf.Serializers.EnumSerializerSByte`1" Activate="Required Public" />
+    <Type Name="ProtoBuf.Serializers.EnumSerializerInt16`1" Activate="Required Public" />
+    <Type Name="ProtoBuf.Serializers.EnumSerializerInt32`1" Activate="Required Public" />
+    <Type Name="ProtoBuf.Serializers.EnumSerializerInt64`1" Activate="Required Public" />
+    <Type Name="ProtoBuf.Serializers.EnumSerializerByte`1" Activate="Required Public" />
+    <Type Name="ProtoBuf.Serializers.EnumSerializerUInt16`1" Activate="Required Public" />
+    <Type Name="ProtoBuf.Serializers.EnumSerializerUInt32`1" Activate="Required Public" />
+    <Type Name="ProtoBuf.Serializers.EnumSerializerUInt64`1" Activate="Required Public" />
+    <Type Name="ProtoBuf.Serializers.ConcurrentDictionarySerializer`3" Activate="Required Public" />
+    <Type Name="ProtoBuf.Serializers.DictionarySerializer`2" Activate="Required Public" />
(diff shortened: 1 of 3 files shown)


Screenbox/Properties/Default.rd.xml
```diff
@@ -21,11 +21,37 @@
...
+    <Type Name="ProtoBuf.Serializers.EnumSerializerUInt64`1" Activate="Required Public" />
+    <Type Name="ProtoBuf.Serializers.ConcurrentDictionarySerializer`3" Activate="Required Public" />
+    <Type Name="ProtoBuf.Serializers.DictionarySerializer`2" Activate="Required Public" />
+    <Type Name="ProtoBuf.Serializers.DictionarySerializer`3" Activate="Required Public" />
+    <Type Name="ProtoBuf.Serializers.DictionaryOfIReadOnlyDictionarySerializer`2" Activate="Required Public" />
+    <Type Name="ProtoBuf.Serializers.ImmutableDictionarySerializer`2" Activate="Required Public" />
+    <Type Name="ProtoBuf.Serializers.ImmutableSortedDictionarySerializer`2" Activate="Required Public" />
+    <Type Name="ProtoBuf.Serializers.ImmutableIDictionarySerializer`2" Activate="Required Public" />
+    <Type Name="ProtoBuf.Serializers.ConcurrentBagSerializer`2" Activate="Required Public" />
+    ...-->
```
United600: 😩 I got tired of adding them individually. I'm not sure we'll see significant savings by continuing.

Screenbox/Properties/Default.rd.xml
```diff
@@ -21,11 +21,37 @@
...
 
+    <!--<Namespace Name="ProtoBuf.Internal" Dynamic="Required All" />-->
+    <Type Name="ProtoBuf.Internal.Level300DefaultSerializer" Dynamic="Required All" />
+    <Type Name="ProtoBuf.Internal.Level300FixedSerializer" Dynamic="Required All" />
+    
+    <Type Name="CommunityToolkit.Mvvm.Messaging.IMessengerExtensions" Dynamic="Required All" />
```
United600: When the app launches, this [method](https://github.com/CommunityToolkit/dotnet/blob/b135626dd54d33b8f05f2ff31591592c004aa848/src/CommunityToolkit.Mvvm/Messaging/IMessengerExtensions.cs#L287) failed even though it is not marked with the [RequiresDynamicCode](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.requiresdynamiccodeattribute?view=net-10.0) attribute.

Maybe we could trim it more efficiently using the [method troubleshooter](https://dotnet.github.io/native/troubleshooter/method.html), though I already attempted it without success.

huynhsontung:
Do we want to enable this before moving away from protobuf?

United600:
> Do we want to enable this before moving away from protobuf?

The gains extend beyond protobuf, but we can hold off for now and use it as test bed. The earlier we start trimming, the better prepared we'll be for NAOT.

United600:
I think this is finally ready. I tested it throughout the entire weekend and didn't run into any issues. Memory usage was even slightly lower, which makes sense now that `Screenbox.dll` is smaller by nearly one third compared to its original size. We should also see a small improvement in launch times.

The only features I didn't test were the network-related ones and Xbox.

PS. I even hit the lowest memory usage so far, measured after several hours of playing music. Though to be fair, I kept the network page open the entire time.

Screenbox task manager memory usage

huynhsontung (approved):

PR #826, opened 2026-03-28
refactor: lively wallpaper feature into a standalone project

United600:
All Lively Wallpaper models, helpers, services, and view models have been moved from `Screenbox.Core` to a new `Screenbox.Lively` project and namespace.
~~Fixes the feature not unloading correctly when selecting back from it.~~

~~The "Default" visualizer stuff can be improved, maybe we could reuse the `DefaultStringConverter` instead of doing the `ResourceLoader` stuff in the ViewModel.~~

The next step will likely involve moving the assets and controls from the `Screenbox` project.


Screenbox/Pages/PlayerPage.xaml
```diff
@@ -264,7 +264,7 @@
             Grid.Row="0"
             Grid.RowSpan="3"
             Grid.Column="0"
-            x:Load="False"
+            x:Load="{x:Bind ViewModel.ShowVisualizer, Mode=OneWay}"
```
huynhsontung: Once the element is loaded, it's invalid to "unload" it. That's why I used `FindName` in the code behind. So that this operation is one way.

United600: Sorry, I'm not sure I understand.

Right now, switching back to the default visualizer still keeps the feature active. That appears to have fixed the issue, and I haven't encountered any problems. I will revert it as it likely doesn't belong in here in the first place.

huynhsontung: I think I missed the switching logic when I reviewed. I saw `x:Bind` and `OneWay` and I assume it would toggle during runtime. I agree that this is out of scope for this PR.

huynhsontung (approved):

PR #836, opened 2026-04-09
chore: fix naming rule and enforce LF line endings for font source files

United600:


huynhsontung:
Why is this necessary? Does using CRLF break the editor config?

United600:
> Why is this necessary? Does using CRLF break the editor config?

Not referring to the .editorconfig, it's about the font source files. FontForge generates UFO source files with LF line endings, which makes checking the diffs annoying to read before commiting.

huynhsontung (approved):

PR #848, opened 2026-04-13
feat: adds ability to delete video file in player

### LINKED ISSUE #847: [Feature Request] Ability to delete active video with button and keyboard shortcut
- DerexXD
- 
Image

Image

Was wondering if there could be the ability to delete the active video, through the typical trash can icon, or by pressing the delete key. A lot of photo / gallery viewer softwares have this and it would be cool for this to have it as well!

After deleting, it could default to the next video in queue or just a blank area.

DereC4:
Closes #847 

Figured I'd make my first PR for a tool I like! Tried to keep style consistent and coding principles similar to how y'all implemented other key functions

Added the appropriate success and fail messages like other dialogues, and uses `DeleteAsync` UWP function like is used for some other existing folder operations in this codebase. 

Before deleting, the current video is moved to the next or prev video and THEN deletion is called.

image

---

Original Issue was:

Image

Was wondering if there could be the ability to delete the active video, through the typical trash can icon, or by pressing the delete key. A lot of photo / gallery viewer softwares have this and it would be cool for this to have it as well!

### DIFF
--- a/Screenbox/Dialogs/DeleteMediaFileDialog.xaml
+++ b/Screenbox/Dialogs/DeleteMediaFileDialog.xaml
@@ -1,0 +1,58 @@
+<ContentDialog
+    x:Class="Screenbox.Dialogs.DeleteMediaFileDialog"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:strings="using:Screenbox.Strings"
+    Title="{strings:Resources Key=DeleteMediaFile}"
+    CloseButtonText="{strings:Resources Key=Cancel}"
+    PrimaryButtonStyle="{StaticResource AccentButtonStyle}"
+    PrimaryButtonText="{strings:Resources Key=Delete}"
+    Style="{StaticResource DefaultContentDialogStyle}"
+    mc:Ignorable="d">
+
+    <ContentDialog.Resources>
+        <ResourceDictionary>
+            <ResourceDictionary.ThemeDictionaries>
+                <ResourceDictionary x:Key="Default">
+                    <SolidColorBrush x:Key="DangerFillColorPrimaryBrush" Color="{StaticResource SystemFillColorCritical}" />
+                    <SolidColorBrush
+                        x:Key="DangerFillColorSecondaryBrush"
+                        Opacity="0.9"
+                        Color="{StaticResource SystemFillColorCritical}" />
+                    <SolidColorBrush
+                        x:Key="DangerFillColorTertiaryBrush"
+                        Opacity="0.8"
+                        Color="{StaticResource SystemFillColorCritical}" />
+
+                    <StaticResource x:Key="AccentButtonBackground" ResourceKey="DangerFillColorPrimaryBrush" />
+                    <StaticResource x:Key="AccentButtonBackgroundPointerOver" ResourceKey="DangerFillColorSecondaryBrush" />
+                    <StaticResource x:Key="AccentButtonBackgroundPressed" ResourceKey="DangerFillColorTertiaryBrush" />
+                </ResourceDictionary>
+                <ResourceDictionary x:Key="Light">
+                    <SolidColorBrush x:Key="DangerFillColorPrimaryBrush" Color="{StaticResource SystemFillColorCritical}" />
+                    <SolidColorBrush
+                        x:Key="DangerFillColorSecondaryBrush"
+                        Opacity="0.9"
+                        Color="{StaticResource SystemFillColorCritical}" />
+                    <SolidColorBrush
+                        x:Key="DangerFillColorTertiaryBrush"
+                        Opacity="0.8"
+                        Color="{StaticResource SystemFillColorCritical}" />
+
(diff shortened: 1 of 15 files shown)


DereC4:
I should add confirmation lol

United600:
I'm unsure about adding the feature directly to the player, would it be more discoverable in the "see more" context menu instead?

If we implement this, it should also appear in the item context menu on library pages.

Screenbox/Strings/en-US/Resources.resw
```diff
@@ -1048,6 +1055,13 @@
     <value>Are you sure you want to delete '{0}' playlist? You cannot undo this action.</value>
     <comment>#Format[String playlistName]</comment>
   </data>
+  <data name="DeleteMediaFile" xml:space="preserve">
+    <value>Delete file</value>
+  </data>
+  <data name="DeleteMediaFileConfirmation" xml:space="preserve">
+    <value>Are you sure you want to delete '{0}'? You cannot undo this action.</value>
```
United600, line 1062: It's being moved to the recycling bin, correct? If so, the "You cannot undo this action" bit isn't necessary.

Screenbox/Dialogs/DeleteMediaFileDialog.xaml
```diff
@@ -0,0 +1,58 @@
...
+                    <StaticResource x:Key="AccentButtonBackgroundPointerOver" ResourceKey="DangerFillColorSecondaryBrush" />
+                    <StaticResource x:Key="AccentButtonBackgroundPressed" ResourceKey="DangerFillColorTertiaryBrush" />
+                </ResourceDictionary>
+                <ResourceDictionary x:Key="HighContrast">
+                    <StaticResource x:Key="AccentButtonBackground" ResourceKey="SystemColorHighlightColorBrush" />
+                    <StaticResource x:Key="AccentButtonBackgroundPointerOver" ResourceKey="SystemColorButtonTextColorBrush" />
+                    <StaticResource x:Key="AccentButtonBackgroundPressed" ResourceKey="SystemColorButtonFaceColorBrush" />
+                </ResourceDictionary>
+            </ResourceDictionary.ThemeDictionaries>
+        </ResourceDictionary>
```
United600, line 54: Likewise, if it's not a destructive action, we can skip this.

United600:
> Before deleting, the current video is moved to the next or prev video and THEN deletion is called.

What if we prevent deleting the currently playing item? It would simplify implementation.

DereC4:
> > Before deleting, the current video is moved to the next or prev video and THEN deletion is called.
> 
> What if we prevent deleting the currently playing item? It would simplify implementation.

Ah that was kind of the whole point though, similar to using a photo gallery app when you want to delete your current media you're looking at, it goes to the next one 

Unless I'm misunderstanding this oops

DereC4:
> I'm unsure about adding the feature directly to the player, would it be more discoverable in the "see more" context menu instead?
> 
> If we implement this, it should also appear in the item context menu on library pages.

Agreed, tho I would still like to have the feature to be in the actiev player (like windows photo viewer below)

image

DereC4:
Image


Kinda like this

Screenbox/Strings/en-US/Resources.resw
United600: So the [StorageDeleteOption](https://learn.microsoft.com/uwp/api/windows.storage.storagedeleteoption) includes a 'permanently delete' value, so we could implement 'Shift+Del' support. If we do then the dialog if fine.

United600:
> Ah that was kind of the whole point though, similar to using a photo gallery app when you want to delete your current media you're looking at, it goes to the next one
> 
> Unless I'm misunderstanding this oops

No worries, it's fine. I was just trying to simplify things.

United600:
> Image
> 
> Kinda like this

Yeah. In my opinion, moving the action to the player context menu is the better choice. Maybe we should also add a shared command so it can be reused across other pages.

DereC4:
> > Image
> > Kinda like this
> 
> Yeah. In my opinion, moving the action to the player context menu is the better choice. Maybe we should also add a shared command so it can be reused across other pages.

Sounds like a plan

The delete key shortcut is fine right ? Seems intuitive to me

DereC4:
Any other code review comments for the time being or implementation tips

United600:
> Any other code review comments for the time being or implementation tips

You could copy the Photos delete dialog behavior and and include a checkbox to remember whether to skip the warning.

> Before deleting, the current video is moved to the next or prev video and THEN deletion is called.

I still don't like all the `deleteIndex` logic, if it's a permission issue could we [stop](https://docs.videolan.me/libvlcsharp/api/LibVLCSharp.Shared.MediaPlayer.html#LibVLCSharp_Shared_MediaPlayer_Stop) the playback instead?

Other than that not really, though we may need to revisit how we currently use the Trash icon in some places.

DereC4:
wheee

DereC4:
can anyone reopen this oops

United600:
> can anyone reopen this oops

Doesn't seem like we can. You can open a new pull request and make sure to fix the build errors. The projects were missing the newly added files.

DereC4:
Damit. I'll try to recover the branch ;-;

huynhsontung:
You can recover the branch on this PR by using GitHub CLI. Then you can fork this repo and push the recovered branch to your own fork.

```
gh pr checkout 848
```

DereC4:
ok i'll try again soon, would rlly like this to become a feature

DereC4:
i did it!!

PR #850, opened 2026-04-15
l10n: Implement localized character grouping support

### LINKED ISSUE #843: The media library grouping list supports sorting by the first letter of Chinese pinyin.

Currently, media items with Chinese names are all grouped under the "..." category, making it difficult to quickly locate and find items.

You can refer to the effects achieved by the built-in media player in Windows 11.

N/A

**Additional context**

Image

Screenbox

Image

Image

Windows Media Player

zuozishi:
This pull request refactors and enhances the media grouping logic throughout the application to use Windows' `CharacterGroupings` for more accurate and locale-aware alphabetical grouping. It replaces custom grouping logic with the system-provided approach, updates related ViewModels to use the new grouping helpers, and improves group label handling in the UI.

Closes #843

### DIFF
--- a/Screenbox.Core/Helpers/MediaGroupingHelpers.cs
+++ b/Screenbox.Core/Helpers/MediaGroupingHelpers.cs
@@ -1,22 +1,45 @@
-using System.Globalization;
+﻿#nullable enable
+
+using System;
+using System.Collections.Generic;
 using System.Linq;
+using Windows.Globalization.Collation;
 
 namespace Screenbox.Core.Helpers;
 
 public static class MediaGroupingHelpers
 {
-    public const string GroupHeaders = "&#ABCDEFGHIJKLMNOPQRSTUVWXYZ\u2026";
-
     public const string OtherGroupSymbol = "\u2026";
 
-    public static string GetFirstLetterGroup(string name)
+    public static readonly IReadOnlyList<string> CharacterGroupLabels;
+
+    public static readonly int MaxGroupLabelLength;
+
+    private static readonly CharacterGroupings _characterGroupings = new();
+
+    private static readonly HashSet<string> _characterGroupSet;
+
+    static MediaGroupingHelpers()
     {
-        if (string.IsNullOrEmpty(name)) return OtherGroupSymbol;
-        char letter = char.ToUpper(name[0], CultureInfo.CurrentCulture);
-        if ("ABCDEFGHIJKLMNOPQRSTUVWXYZ".Contains(letter))
-            return letter.ToString();
-        if (char.IsNumber(letter)) return "#";
-        if (char.IsSymbol(letter) || char.IsPunctuation(letter) || char.IsSeparator(letter)) return "&";
+        CharacterGroupLabels = [.. _characterGroupings
+            .Select(x => string.IsNullOrEmpty(x.Label) ? OtherGroupSymbol : x.Label)
+            .Distinct()];
+        MaxGroupLabelLength = CharacterGroupLabels.Select(x => x.Length).Max();
+
+        _characterGroupSet = new HashSet<string>(CharacterGroupLabels, StringComparer.Ordinal);
+    }
+
+    public static string GetCharacterGroupLabel(string name)
+    {
+        try
+        {
+            string? label = _characterGroupings.Lookup(name);
+            if (string.IsNullOrEmpty(label) || !_characterGroupSet.Contains(label))
+                return OtherGroupSymbol;
+
+            return label;
+        }
+        catch { }
         return OtherGroupSymbol;
     }
 }
--- a/Screenbox.Core/Helpers/MediaGroupingHelpers.cs
+++ b/Screenbox.Core/Helpers/MediaGroupingHelpers.cs
@@ -3,6 +3,8 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Screenbox.Core.Services;
+using Windows.Globalization;
 using Windows.Globalization.Collation;
 
 namespace Screenbox.Core.Helpers;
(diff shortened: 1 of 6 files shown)


Screenbox/Behaviors/GroupingOverviewBehavior.cs
```diff
@@ -52,11 +55,13 @@ private void UpdateGroupViewItemWidth()
...
             var element = (FrameworkElement)child;
-            element.Width = GroupType == "year"
-                ? 80
-                : AssociatedObject.HorizontalAlignment != HorizontalAlignment.Stretch
-                    ? double.NaN
-                    : itemWidth;
+            element.Width = GroupType switch
+            {
+                "year" => 80,
+                null or "" when MediaGroupingHelpers.MaxGroupLabelLength > 1 => 100,
```
huynhsontung, line 61: When the max label length is more than 1 character, we hardcode the width to 100? Is this the best way to handle this? Perhaps we can estimate the label's actual width from its length? How about just leaving it to `NaN` for the gridview to handle?

Screenbox.Core/Helpers/MediaGroupingHelpers.cs
```diff
@@ -1,22 +1,54 @@
...
     {
-        if (string.IsNullOrEmpty(name)) return OtherGroupSymbol;
-        char letter = char.ToUpper(name[0], CultureInfo.CurrentCulture);
-        if ("ABCDEFGHIJKLMNOPQRSTUVWXYZ".Contains(letter))
-            return letter.ToString();
-        if (char.IsNumber(letter)) return "#";
-        if (char.IsSymbol(letter) || char.IsPunctuation(letter) || char.IsSeparator(letter)) return "&";
+        _characterGroupings = string.IsNullOrWhiteSpace(ApplicationLanguages.PrimaryLanguageOverride)
+            ? new CharacterGroupings()
+            : new CharacterGroupings(ApplicationLanguages.PrimaryLanguageOverride);
```
United600, line 27:
```suggestion
        _characterGroupings = new CharacterGroupings(CultureInfo.CurrentCulture.Name);
```

We could skip this check and simply pass the culture directly.

Screenbox.Core/Helpers/MediaGroupingHelpers.cs
```diff
@@ -1,22 +1,54 @@
...
 public static class MediaGroupingHelpers
 {
-    public const string GroupHeaders = "&#ABCDEFGHIJKLMNOPQRSTUVWXYZ\u2026";
-
     public const string OtherGroupSymbol = "\u2026";
 
-    public static string GetFirstLetterGroup(string name)
+    public static IReadOnlyList<string> CharacterGroupLabels { get; }
+
+    public static int MaxGroupLabelLength { get; }
```
United600, line 18: AFAICS, only Chinese uses labels with multiple characters, so we could move this to the behavior and implement a check like this:

https://github.com/huynhsontung/Screenbox/blob/a455c314dd1678c2bc7d7c4b80ab4877b7d45225/Screenbox/Extensions/AcceleratorService.cs#L39

```csharp
private static readonly bool _isChineseCulture = CultureInfo.CurrentCulture.TwoLetterISOLanguageName is "zh";

...

null or "" when _isChineseCulture => 100,
```

Screenbox.Core/Helpers/MediaGroupingHelpers.cs
```diff
@@ -1,22 +1,54 @@
...
+
+    public static string GetCharacterGroupLabel(string name)
+    {
+        try
+        {
+            string? label = _characterGroupings.Lookup(name);
+            if (string.IsNullOrEmpty(label) || !_characterGroupSet.Contains(label))
+                return OtherGroupSymbol;
+
+            return label;
```
United600:
```suggestion
            return string.IsNullOrEmpty(label) || !_characterGroupSet.Contains(label)
                ? OtherGroupSymbol
                : label;
```

United600:
@zuozishi Do you know why there is such a difference between the supported Chinese locales? Is it because of the names I used?

| Chinese (Simplified) | Chinese (Traditional) |
|----------------------|-----------------------|
| Songs page with zh-Hans group headers | Songs page with zh-Hant group headers |

In any case, it works great in all locales.

https://github.com/user-attachments/assets/6732b36a-1714-4e32-86a3-4b6daa880806

Songs page with japanese group headers

Screenbox/Behaviors/GroupingOverviewBehavior.cs
United600: Yeah, we shouldn't set a fixed width at all (MinWidth preferably). But IIRC, without it, the `GridView` would default to the width of the first item and apply it to the rest, which could lead to text trimming.

But we can address this and the screen reader name later.

Screenbox.Core/Helpers/MediaGroupingHelpers.cs
huynhsontung: This is mainly a null check since I don't know if this value is null when you install the app for the first time.

Screenbox.Core/Helpers/MediaGroupingHelpers.cs
huynhsontung: And Japanese :) This method is still just an assumption.

image

United600: 🫠 Ouch, yeah I even have a screenshot of it.

Screenbox.Core/Helpers/MediaGroupingHelpers.cs
United600: It shouldn't, but checking won't hurt.

United600 (approved):

zuozishi:
@United600 Typically, in Chinese applications, text grouping shares a single Label with English A-Z. As for why Microsoft added the two-character prefix "拼音" (pinyin), that's something to @Microsoft. 😀

huynhsontung (approved):
LGTM. Thank you!

PR #860, opened 2026-04-24
feat: add multi-select support to home page recent list

### LINKED ISSUE #853: Add a "Clear recent media history" button on Home page

Not being able to clear recent media history from **Home** page

Add a "Clear recent media history" button on Home page

The **Clear** button is already available in Settings > Privacy

Image

United600:
Closes #853.

https://github.com/user-attachments/assets/d0fbf42e-3773-4441-8e42-1b9933a9d38f

https://github.com/user-attachments/assets/c965895d-03e8-4254-bdab-628d3ffbddb8

### DIFF
--- a/Screenbox.Core/ViewModels/HomePageViewModel.cs
+++ b/Screenbox.Core/ViewModels/HomePageViewModel.cs
@@ -24,6 +24,16 @@ public sealed partial class HomePageViewModel : ObservableRecipient,
 {
     public ObservableCollection<MediaViewModel> Recent { get; }
 
+    [ObservableProperty]
+    [NotifyCanExecuteChangedFor(nameof(PlaySelectedCommand))]
+    [NotifyCanExecuteChangedFor(nameof(PlaySelectedNextCommand))]
+    [NotifyCanExecuteChangedFor(nameof(RemoveSelectedCommand))]
+    private int _selectionCount;
+
+    [ObservableProperty] private bool? _selectionCheckState;
+    [ObservableProperty] private bool _enableMultiSelect;
+    [ObservableProperty] private MediaViewModel? _selectedItemToAdd;
+
     private readonly MediaViewModelFactory _mediaFactory;
     private readonly IFilesService _filesService;
     private readonly ISettingsService _settingsService;
@@ -41,6 +51,7 @@ public sealed partial class HomePageViewModel : ObservableRecipient,
         _changeDebounceTimer = _dispatcherQueue.CreateTimer();
         _pathToMruMappings = new Dictionary<string, string>();
         Recent = new ObservableCollection<MediaViewModel>();
+        SelectionCheckState = GetSelectionCheckState(_selectionCount);
 
         // Activate the view model's messenger
         IsActive = true;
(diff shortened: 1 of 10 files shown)


Screenbox.Core/ViewModels/HomePageViewModel.cs
```diff
@@ -254,4 +333,23 @@ private async Task OpenFolderAsync()
...
+
+    partial void OnSelectionCountChanged(int value)
+    {
+        SelectionCheckState = GetSelectionCheckState(value);
+    }
+
+    partial void OnEnableMultiSelectChanged(bool value)
+    {
+        if (!value) SelectionCount = 0;
+    }
```
United600, line 347: We should move this duplicate code and properties to a shared `SelectionContext`.

Screenbox/Pages/HomePage.xaml
```diff
@@ -166,50 +180,188 @@
...
+                            </MenuFlyout>
+                        </AppBarButton.Flyout>
+                    </AppBarButton>
+                    <AppBarButton
+                        Command="{x:Bind ViewModel.RemoveSelectedCommand}"
+                        CommandParameter="{x:Bind RecentFilesGridView.SelectedItems}"
+                        Icon="Clear"
+                        KeyboardAcceleratorTextOverride="{x:Bind strings:KeyboardResources.VirtualKeyDelete}"
+                        Label="{x:Bind strings:Resources.Remove}" />
+
```
United600, line 259: Initially, I included "Open artist" and "Open album" buttons, but the visibility changes was causing issues, such as the "see more" button appearing even when the overflow menu was empty. We can add those later, along with the "Properties" button.

Screenbox/Controls/PlayQueueControl.xaml.cs
```diff
@@ -128,16 +131,13 @@ private void PlayQueue_OnUnloaded(object sender, RoutedEventArgs e)
...
         {
-            if (_selectionCommand.CanToggleSelection(PlaylistListView))
-            {
-                ViewModel.EnableMultiSelect = true;
-                _selectionCommand.ToggleSelection(PlaylistListView);
-                args.Handled = true;
-            }
+            ViewModel.EnableMultiSelect = true;
+            _selectionCommand.ToggleSelection(PlaylistListView);
+            args.Handled = true;
```
United600, line 140: Should we do this instead?

```csharp
if (!ViewModel.HasItems) return;

var cmd = (ICommand)Application.Current.Resources["SelectDeselectAllCommand"];
var parameter = PlaylistListView;
if (cmd.CanExecute(parameter))
{
    ViewModel.EnableMultiSelect = true;
    cmd.Execute(parameter);
    args.Handled = true;
}
```

Screenbox.Core/ViewModels/HomePageViewModel.cs
huynhsontung: Agree. What should be the lifetime of a `SelectionContext`? Should a new instance of `SelectionContext` be created on every page? I think that would make the most sense.

Screenbox/Controls/PlayQueueControl.xaml.cs
huynhsontung: The current version uses strongly typed functions. I prefer typed parameters to generic parameters with `CanExecute(object)` and `Execute(object)`.

Screenbox.Core/ViewModels/HomePageViewModel.cs
United600: I did some prototyping and was wondering if it's okay to move the `partial void On...Changed(...)` methods. Also, I'm having trouble finding a clean way to get the selection count there.

Could we address that in a separate PR where we merge the `PlayQueueViewModel` and `HomePageViewModel` implementations?

huynhsontung: I agree to keep this PR as-is and implement further changes in a separate PR. For the selection context, since the logic interacts directly with the controls, it doesn't seem to use any properties in `HomePageViewModel`. We can create a new ViewModel just for selection, `SelectionViewModel` perhaps? Contexts are only supposed to have data, similar to models, but observable.

United600: > We can create a new ViewModel just for selection, `SelectionViewModel` perhaps? Contexts are only supposed to have data, similar to models, but observable.

Yeah, that sounds more reasonable.

> I agree to keep this PR as-is and implement further changes in a separate PR.

I'll also start to investigate the best way to consolidate the selection control into a reusable component.

huynhsontung (approved):

PR #868, opened 2026-05-02
fix: playlist items screen reader name

United600:
Used the opportunity to streamline some of our accessibility logic.


United600:
Why did we make this a double?

https://github.com/huynhsontung/Screenbox/blob/ed26ed1d0de53da6d64503cf719d04f58fd5473a/Screenbox.Core/ViewModels/PlaylistViewModel.cs#L23

huynhsontung:
> Why did we make this a double?

For binding, whatever control is bound to this property expects a double value.

United600:
> > Why did we make this a double?
> 
> 
> 
> For binding, whatever control is bound to this property expects a double value.
> 
> 
> 
> 

Makes sense. We could use x:Bind casting and make it a Int/Uint, right?

huynhsontung:
Yes, but there's no point doing so. Int and Uint can be assigned to double and that's what the control uses

PR #870, opened 2026-05-03
refactor: selection logic into shared ViewModel

United600:


Screenbox.Core/ViewModels/HomePageViewModel.cs
```diff
@@ -76,6 +74,17 @@ public async void OnLoaded()
...
+    private void Selection_OnPropertyChanged(object sender, PropertyChangedEventArgs e)
+    {
+        if (e.PropertyName == nameof(SelectionViewModel.SelectedItemCount))
+        {
+            PlaySelectedCommand.NotifyCanExecuteChanged();
+            PlaySelectedNextCommand.NotifyCanExecuteChanged();
+            AddSelectedToQueueCommand.NotifyCanExecuteChanged();
+            RemoveSelectedCommand.NotifyCanExecuteChanged();
+        }
+    }
```
United600, line 86: Should we add a `SelectedItemCountChanged` event on the `ViewModel` instead of relying on the `PropertyChanged` approach.

huynhsontung: I don't really see the benefits. We already use `PropertyChanged` everywhere, so let's keep it consistent.

United600: What about now that I removed the `SelectedItemCount` observable property? 
I'm using `IsAllSelected` and it works fine. However, for better correctness, an event might be a more appropriate solution.

Screenbox.Core/ViewModels/SelectionViewModel.cs
```diff
@@ -0,0 +1,131 @@
...
+
+    /// <summary>
+    /// Sets the source collection for selection and updates the selection state.
+    /// </summary>
+    /// <param name="source">A collection of items to be used as the selection source.</param>
+    public void SetItemsSource(IReadOnlyCollection<object>? source)
+    {
+        if (_sourceCollection is INotifyCollectionChanged oldCollection)
+        {
+            oldCollection.CollectionChanged -= SourceCollection_OnCollectionChanged;
```
United600, line 70: This is never called, should I be concerned about memory leaks? The same applies to `SelectedItems_OnCollectionChanged`?

huynhsontung: I actually don't know. The set source usage in `PlayQueueViewModel` may be problematic because the lifetime of `MediaListViewModel` playlist is the whole app session.  Maybe we can use `WeakEventListener`? Or set the selection view model's source to null on unload, but this is unreliable.

United600: How about implementing `IDisposable` and invoking `Dispose()` when the view is `Unloaded`/`OnNavigatedFrom`?

> Maybe we can use WeakEventListener?

I thought about it and began implementing, but then discovered we're missing the `CommunityToolkit.Uwp.Helpers` in the Core project.

huynhsontung: Using `IDisposable` is like a last ditch effort. I want to avoid using `IDisposable` as much as possible. How about we use [ReadOnlyObservableCollection](https://learn.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.readonlyobservablecollection-1?view=netframework-4.8.1&viewFallbackFrom=net-10.0). Every time we set source, we create a new instance of `ReadOnlyObservableCollection`. The lifetime of this object will be known within the `SelectionViewModel`. I trust system class to handle lifecycle better properly.

United600: I attempted to use [WeakReference](https://learn.microsoft.com/en-us/dotnet/api/system.weakreference) since I'm more familiar with it, but couldn't get the GC to collect it. Perhaps my implementation was incorrect, and I'm struggling to apply your suggestion. I'm not very familiar with managing ViewModel lifetimes, and I might be debugging it the wrong way.

huynhsontung: Never mind, `ReadOnlyObservableCollection.CollectionChanged` in UWP for some reason is a protected event.

huynhsontung: Let's ignore this issue for now. The owner class `PlayQueueViewModel` also violates this rule. Something to resolve later. I need to break `MediaListViewModel` into a context and a view model or something else.

huynhsontung (approved):

PR #887, opened 2026-05-14
feat: add configurable seek step intervals

United600:
Adds settings for rewind and fast forward step intervals, allowing users to customize the number of seconds skipped during seek actions.

**Normal**
Seek step settings

**Expanded**
Seek step settings with the options allowed expanded


Screenbox/Pages/MainPage.xaml.cs
```diff
@@ -73,7 +73,7 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
 
         protected override void OnKeyDown(KeyRoutedEventArgs e)
         {
-            e.Handled = ViewModel.ProcessGamepadKeyDown(e.Key);
+            ViewModel.ProcessGamepadKeyDown(e.Key);
```
huynhsontung, line 76: Why do we stop setting the handled property here? Is this a breaking change? Is there any handler downstream of this that we want to accommodate?

United600: It would break gamepad navigation, unsure if it caused other problems. Didn't experienced any regressions after removing it.

United600: https://github.com/huynhsontung/Screenbox/blob/0a3cacff414b73bbf2a729197a7e5cf0d5a87350/Screenbox.Core/ViewModels/MainPageViewModel.cs#L142-L143

Maybe if we remove this early return we can restore the `e.Handled = true`?

United600: > Maybe if we remove this early return we can restore the `e.Handled = true`?

Negative, it breaks the gamepad navigation 😑.

huynhsontung (approved):

PR #889, opened 2026-05-16
refactor: search suggestion model and templates

United600:
Removes search box flickering to improve user experience. Rebuilds the suggestions collection using equality checks rather than clearing and rebuilding it on every query change.
Replaces the `SearchSuggestionConverter.cs` with a `DataTemplateSelector` to handle template modifications, and moves the `GridViewItemTemplates.xaml` to the new `Templates` folder.

I believe the search implementation could benefit from more refactoring, we might be handling too much logic in the `MainPageViewModel`.

**Previous**

https://github.com/user-attachments/assets/96f02c46-e9a1-4431-a535-c75ce09eef4a

**New**

https://github.com/user-attachments/assets/3f2e9779-9dc3-4a80-9454-ad0b4a2106dc


Screenbox.Core/Models/SearchSuggestion.cs
```diff
@@ -0,0 +1,136 @@
...
+using System;
+using System.Runtime.CompilerServices;
+using Screenbox.Core.Enums;
+
+namespace Screenbox.Core.Models;
+
+/// <summary>
+/// Represents a search suggestion item with a type, display text, and associated data.
+/// </summary>
+public sealed class SearchSuggestion : IEquatable<SearchSuggestion>
```
huynhsontung: I think if you make this a record, you get all the comparison operator overload, `IEquatable` for free, no need to overload them 😅 

```suggestion
public sealed record SearchSuggestion
```

Screenbox.Core/ViewModels/MainPageViewModel.cs
```diff
@@ -147,27 +155,52 @@ public void OnDrop(DataPackageView data)
...
-                SearchSuggestions.Add(new SearchSuggestionItem(searchQuery, null, SearchSuggestionKind.None));
+                SearchSuggestions.Add(newSuggestions[i]);
             }
         }
+
+        // Remove any extra old suggestions that are no longer present.
+        while (SearchSuggestions.Count > newSuggestions.Count)
+        {
+            SearchSuggestions.RemoveAt(SearchSuggestions.Count - 1);
+        }
```
huynhsontung: We already have an extension for syncing collections. This whole block can be just

```cs
SearchSuggestions.SyncItems(newSuggestions);
```

Screenbox.Core/Models/SearchSuggestion.cs
United600: Oh! The handling of the `Type` can be a bit tricky, but I'll test it to see if we can switch to a record.

Screenbox.Core/ViewModels/MainPageViewModel.cs
United600: 😲 I'll also replace that entire block, though getting the `Type` right was a bit tricky.

United600: I haven't switched over yet, but if you believe a minor regression in the handling of 'No results' doesn't justify a custom solution, I can replace it with the `SyncItems` extension. There also seems to be a slight performance drop, but it may be negligible.

PR #895, opened 2026-05-19
feat: adds ability to delete video file in player (2nd attempt)

### LINKED ISSUE #847: [Feature Request] Ability to delete active video with button and keyboard shortcut

DereC4:
Recovered PR from #848 

Closes #847 

- Figured I'd make my first PR for a tool I like! Tried to keep style consistent and coding principles similar to how y'all implemented other key functions
- Added the appropriate success and fail messages like other dialogues, and uses `DeleteAsync` UWP function like is used for some other existing folder operations in this codebase. 
- ~~I want the behavior to be after deletion, you move onto the next video in your list, instead of it just turning gray, so that smooth deletion can happen similar to any photo gallery or video gallery on your phone. That way people don't have to re-open their next video on every delete and can smoothly delete one by one as they scroll through a video, then next video~~ (I have failed to do this )
- This PR might also be exposing a thing where we dont add drag-drop files properly, leading them to still have the read flag and be treated differently from files that were opened with the app. That is why there's a failure to delete them. I will look into that

image

image

image

---

Original Issue was:

Image

Was wondering if there could be the ability to delete the active video, through the typical trash can icon, or by pressing the delete key. A lot of photo / gallery viewer softwares have this and it would be cool for this to have it as well!


DereC4:
thanks. I'm on new apartment wifi right now so i havent been able to redownload visual studio but ill get to it sometime. Hope to get this added soon

DereC4:
image


Added pics

DereC4:
image

DereC4:
image


Delete works

DereC4:
hmm ran into a 
```

Exception thrown: 'System.Exception' in System.Private.CoreLib.dll
WinRT information: This file is restricted to read access and may not be modified or deleted.
The thread 5300 has exited with code 0 (0x0).
```

DereC4:
nvm im struggling sorry. Trying to figure out why the heck i cant delete a file now

DereC4:
@huynhsontung any thoughts? I'm at my wits end right now, ive been trying to debug for a while why some cases it'll show the file as read only despite all my attempts to remove read only flag, or add a delay before deleting. 

Assumed by pausing playback the file could then be deleted

cc @United600

United600:
I peeked, and it appears that drag-and-drop files into the app breaks the feature. Using the [file or folder picker UI](https://learn.microsoft.com/en-us/windows/uwp/files/quickstart-using-file-and-folder-pickers) or launching a file via `Windows Explorer` works fine, maybe we aren't adding the files to the [FutureAccessList](https://learn.microsoft.com/en-us/windows/apps/develop/files/track-recently-used-files-folders)? That would explain why we don't have the necessary permissions.

Let's try to address the root cause first before adding broad filesystem capability. I'd also suggest a significant refactor, as the current implementation is unnecessarily complex. If it helps, you could remove the "remember permanent delete" option entirely. I think we can streamline this by using a `RelayCommand`, eliminating the need for the `DeleteCurrentMediaFileRequestMessage.cs`.

DereC4:
> I peeked, and it appears that drag-and-drop files into the app breaks the feature. Using the [file or folder picker UI](https://learn.microsoft.com/en-us/windows/uwp/files/quickstart-using-file-and-folder-pickers) or launching a file via `Windows Explorer` works fine, maybe we aren't adding the files to the [FutureAccessList](https://learn.microsoft.com/en-us/windows/apps/develop/files/track-recently-used-files-folders)? That would explain why we don't have the necessary permissions.
> 
> Let's try to address the root cause first before adding broad filesystem capability. I'd also suggest a significant refactor, as the current implementation is unnecessarily complex. If it helps, you could remove the "remember permanent delete" option entirely. I think we can streamline this by using a `RelayCommand`, eliminating the need for the `DeleteCurrentMediaFileRequestMessage.cs`.

ok

DereC4:
You mean the remember delete thing right 
oop yes u do

DereC4:
I'll get to the refactor first

DereC4:
Yep deleting in normal usage works smoothly! I now want to make the behavior go to the next video like a photo gallery does, so you can smoothly delete clips you're browsing in a folder instead of having to reopen the next video

DereC4:
After testing out the user experience i think doing it this way will save a ton of time and having to re-open video for many users

DereC4:
Possibly unrelated, i started getting this error with opening some video files that were working previously, under this debug branch and idk what i did 😭

image

Gonna revert to prev commit to check

DereC4:
image

Still happens. I may have broken something. Going one step back

DereC4:
Fixed. It was indeed me. Now we know

DereC4:
Ok I think it is good to go now, at least for the delete functions itself, and I removed the confirmation check to clean things up. 

This might also be exposing a thing where we dont add drag-drop files properly, leading them to still have the read flag and be treated differently from files that were opened with the app.

DereC4:
I am not sure why https://github.com/huynhsontung/Screenbox/pull/895/changes/a1d449dac8b39cf66a616b9e663cfb5b501b97b9 caused a memory error when I tried opening video files

DereC4:
Ok oop ima leave it like this for now, so when I tried using `GetNextAfterDeleteAsync` like how we do in the codebase to get to the next video, it started causing the crashes shown above (but that happened when i would open the video file, not even getting to the delete part yet)

I am not sure why, if anyone could shed some insight. Heading off for now. 

(Also the drag and drop thing was indeed a separate issue you are correct)

Screenbox/Pages/PlayerPage.xaml
```diff
@@ -471,6 +472,7 @@
 
         <controls:PlayerControls
             x:Name="PlayerControls"
+            DeleteMediaFileRequested="PlayerControls_OnDeleteMediaFileRequested"
```
huynhsontung, line 475: We try to avoid creating events in the view layer, especially between controls. All communication should happen between the view models.

Screenbox/Pages/PlayerPage.xaml.cs
```diff
@@ -523,6 +527,51 @@ private void SeekToPercentageKeyboardAccelerator_OnInvoked(KeyboardAccelerator s
...
     }
 
+    private async void PlayerControls_OnDeleteMediaFileRequested(object sender, RoutedEventArgs e)
+    {
+        await TryDeleteCurrentMediaFileAsync(VirtualKey.Delete, VirtualKeyModifiers.None);
+    }
+
+    private async Task<bool> TryDeleteCurrentMediaFileAsync(VirtualKey key, VirtualKeyModifiers modifiers)
+    {
+        if (Interlocked.Exchange(ref _deleteDialogOpen, 1) == 1)
```
huynhsontung, line 537: Huh, I didn't know this existed. Neat! But probably unnecessary as this will always run on the UI thread.

Screenbox.Core/ViewModels/PlayerPageViewModel.cs
```diff
@@ -592,6 +592,35 @@ public bool ProcessFrameSteppingKeyDown(VirtualKey key)
...
+    /// <param name="modifiers">The modifier keys held during the key press.</param>
+    /// <returns><see langword="true"/> if handled; otherwise, <see langword="false"/>.</returns>
+    public bool ProcessDeleteKeyDown(VirtualKey key, VirtualKeyModifiers modifiers)
+    {
+        if (PlayerVisibility != PlayerVisibilityState.Visible)
+        {
+            return false;
+        }
+
+        if (key != VirtualKey.Delete || modifiers != VirtualKeyModifiers.None)
```
huynhsontung, line 608: If the method is already called `ProcessDeleteKeyDown` why check for the Delete key here? Shouldn't the caller check this before calling this method?

Screenbox.Core/ViewModels/PlayerPageViewModel.cs
```diff
@@ -592,6 +592,35 @@ public bool ProcessFrameSteppingKeyDown(VirtualKey key)
...
+        {
+            return false;
+        }
+
+        try
+        {
+            _ = Messenger.Send(new DeleteCurrentMediaFileRequestMessage());
```
huynhsontung, line 615: Ooh async task discard. Is there a reason for this? Why return a task in the first place if we just fire-and-forget here?

Screenbox.Core/ViewModels/MediaListViewModel.cs
```diff
@@ -620,6 +626,62 @@ private void CreatePlaylistAndPlay(MediaViewModel nextItem)
...
+            }
+            else
+            {
+                MediaPlayer?.Pause();
+            }
+        }
+        catch (Exception)
+        {
+            return Fail();
+        }
```
huynhsontung, line 670: Not the standard way to stop playback. A better way is to set `CurrentItem = null` or set to another item depending on the desired behavior.

Screenbox/Pages/PlayerPage.xaml.cs
```diff
@@ -523,6 +527,51 @@ private void SeekToPercentageKeyboardAccelerator_OnInvoked(KeyboardAccelerator s
         args.Handled = ViewModel.ProcessPercentJumpKeyDown(args.KeyboardAccelerator.Key);
     }
 
+    private async void PlayerControls_OnDeleteMediaFileRequested(object sender, RoutedEventArgs e)
+    {
+        await TryDeleteCurrentMediaFileAsync(VirtualKey.Delete, VirtualKeyModifiers.None);
```
huynhsontung, line 532: So the only reason we have an event on `PlayerControls` is to use the `TryDeleteCurrentMediaFileAsync` method? Why does this have to be called on the `PlayerPage`?

huynhsontung:
Also, please lint your XAML files using XAML Styler.

PR #904, opened 2026-05-23
refactor: minor improvements to templated controls

United600:
Some improvements to code readability and maintainability. Updates dependency properties with `typeof(Enum/Struct)` to align with [best practices](https://learn.microsoft.com/en-us/windows/apps/develop/platform/xaml/custom-dependency-properties#property-changed-behavior-for-structures-and-enumerations).

**ContentUnavailableView**
- Extracts notification string logic to a new `GetNotificationString()` method
- Adds the null-forgiving operator to the Title property in the `GetNotificationString()` return statement
- Separates theme resources into new `ResourceDictionary` for improved maintainability

**CommadBarEx**
- Reorders unsubscription to ensure `_expandButton` is disposed before unsubscribing from the method

**TitleBar**
- Improves reset method to also clear custom caption button colors
- Renames some public methods for improved clarity
- Refactors `HeightMode` dependency property changed callback
- Fixes content changes not triggering the correct visual state
- Moves cached window/title bar references to the constructor
- Separates theme resources into new `ResourceDictionary` for improved maintainability

**NavigationViewEx**
- Refactors `ContentVisibility` and `ContentTranslationDirection` dependency property callbacks
- Extracts built-in buttons initialization into helper methods
- Updates null checks to modern C# style
- Improves docs

### DIFF
--- a/Screenbox/Controls/CommandBarEx/CommandBarEx.cs
+++ b/Screenbox/Controls/CommandBarEx/CommandBarEx.cs
@@ -68,10 +68,11 @@ public sealed class CommandBarEx : CommandBar
         {
             _expandButton = expandButton;
 
-            var visual = ElementCompositionPreview.GetElementVisual(expandButton);
-            var compositor = visual.Compositor;
+            var buttonVisual = ElementCompositionPreview.GetElementVisual(expandButton);
+            var compositor = buttonVisual.Compositor;
+            var animationGroup = CreateShowAnimationGroup(compositor);
             ElementCompositionPreview.SetIsTranslationEnabled(expandButton, true);
-            ElementCompositionPreview.SetImplicitShowAnimation(expandButton, CreateShowAnimationGroup(compositor));
+            ElementCompositionPreview.SetImplicitShowAnimation(expandButton, animationGroup);
 
             UpdateMoreButtonStyle();
         }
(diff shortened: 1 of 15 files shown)


Screenbox/Controls/TitleBar/TitleBar.cs
```diff
@@ -370,22 +412,21 @@ private void UpdateFooter()
...
-
-        titleBar.ButtonForegroundColor = GetNullableColorFromBrush(CaptionButtonForegroundBrush);
-        titleBar.ButtonHoverForegroundColor = GetNullableColorFromBrush(CaptionButtonForegroundPointerOverBrush);
-        titleBar.ButtonPressedForegroundColor = GetNullableColorFromBrush(CaptionButtonForegroundPressedBrush);
-        titleBar.ButtonInactiveForegroundColor = GetNullableColorFromBrush(CaptionButtonForegroundInactiveBrush);
+        // Skip updates when not visible to prevent overwriting colors
+        // set by another TitleBar instance that is currently active.
+        if (_applicationViewTitleBar is null || Visibility == Visibility.Collapsed) return;
+
+        _applicationViewTitleBar.ButtonBackgroundColor = GetNullableColorFromBrush(CaptionButtonBackgroundBrush);
```
United600:
```suggestion
        _applicationViewTitleBar.ButtonBackgroundColor = (CaptionButtonBackgroundBrush as SolidColorBrush)?.Color;
```

We could use the `as` operator and remove the `GetNullableColorFromBrush()` method.

Screenbox/Controls/CommandBarEx/CommandBarEx.cs
```diff
@@ -89,4 +98,34 @@ private void UpdateMoreButtonStyle()
...
+        opacityAnimation.InsertKeyFrame(1f, 1f, linearEasingFunction);
+        opacityAnimation.Duration = TimeSpan.FromMilliseconds(83);
+        opacityAnimation.Target = "Opacity";
+
+        var animationGroup = compositor.CreateAnimationGroup();
+        animationGroup.Add(translationAnimation);
+        animationGroup.Add(opacityAnimation);
+
+        return animationGroup;
+    }
```
huynhsontung: I'm not sure how to feel about this. I don't think this is cleaner than the XAML we currently have.

United600: Yeah, I began with a small scope because, eventually, I plan to migrate the `NavigationViewEx` animations to this approach. That way, we can finally resolve a long-standing bug in the current implementation.

I guess that when I address that, I can create an `AnimationsHelper` with generic methods that handle both cases.

Screenbox/Controls/TitleBar/TitleBar.cs
huynhsontung: Agree. Using `as` is more explicit.

PR #906, opened 2026-05-23
chore: remove UTF-8 with BOM, use UTF-8 encoding instead

huynhsontung:
UTF-8 with BOM encoding breaks Crowdin syncing. It's better to follow the cross-platform standard and use UTF-8 instead of going the Windows-only route.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,6 +6,7 @@ root = true
 indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
+charset = utf-8
 
 # Solution files
 [*.sln]
@@ -34,7 +35,6 @@ indent_size = 2
 
 # XML user interface files
 [*.xaml]
-charset = utf-8-bom
 indent_size = 4
 
 # Data serialization files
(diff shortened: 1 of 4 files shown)


huynhsontung:
This will trigger a re-encoding on every file on save in Visual Studio. We can re-encode everything in a single PR or re-encode as we go.

United600:
> UTF-8 with BOM encoding breaks Crowdin syncing

Why is Crowdin even checking the encoding of XAML and C# source files? And the line endings of the resources files? What a mess.

> It's better to follow the cross-platform standard and use UTF-8 instead of going the Windows-only route.

No concerns from me, hopefully we avoid any subtle bugs. It's odd that even in UWP NAOT, XAML files still don't default to UTF-8, unlike WinUI3.

United600:
> We can re-encode everything in a single PR or re-encode as we go.

Like with the `.glif` files, it's probably better as we go, since we have open PRs that would undo any changes.

huynhsontung:
> Why is Crowdin even checking the encoding of XAML and C# source files? And the line endings of the resources files?

This is my guess. Looking at their code, it shouldn't trigger a new commit every sync but it does. When comparing commits I see no changes so I assume it's because of line endings and file encoding. The action is running inside a Linux container after all. Plus having a common encoding would be better.

huynhsontung:
Also Copilot generates .cs files with UTF-8 encoding lol

PR #917, opened 2026-05-28
chore: replace strings:Resources markups with x:Bind where possible

huynhsontung:
UWP Native AOT will not support the `strings:Resources` markup anymore.

### DIFF
--- a/Screenbox/App.xaml
+++ b/Screenbox/App.xaml
@@ -67,11 +67,11 @@
                         TrueValue="True" />
                     <ctConverters:BoolToObjectConverter
                         x:Key="BoolToPlayPauseTextConverter"
-                        FalseValue="{strings:Resources Key=Play}"
-                        TrueValue="{strings:Resources Key=Pause}" />
-                    <converters:DefaultStringConverter x:Key="GenreTextConverter" Default="{strings:Resources Key=UnknownGenre}" />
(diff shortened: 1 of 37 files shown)


United600:
> UWP Native AOT will not support the `strings:Resources` markup anymore.

What about as a parameter? Something like this:

```xml
<Control Title="{x:Bind converters:DefaultStringConverter.GetValueOrFallback(Name, strings:Resources.UnknownArtist), Mode=OneWay}" />
```

If that's the case, it's unfortunate. I was counting on it for upcoming plans. Looks like I'll have to drop x:Bind function in favor of using `Converter` for this use case 😩.

huynhsontung:
> > UWP Native AOT will not support the `strings:Resources` markup anymore.
> 
> What about as a parameter? Something like this:
> 
> ```xml
> <Control Title="{x:Bind converters:DefaultStringConverter.GetValueOrFallback(Name, strings:Resources.UnknownArtist), Mode=OneWay}" />
> ```

Yes, this will still work. It's not using the `strings: Resources` markup. It's property binding.

huynhsontung:
> UWP Native AOT will not support the strings:Resources markup anymore.

Found out that this was false. For some reason, only **generated** `MarkupExtension` doesn't work. If I copy the same extension and paste locally, then it works. 🤷

PR #926, opened 2026-06-06
fix: ReswPlus markup extension not working with Native AOT

huynhsontung:


huynhsontung:
cc @United600 😩

United600:
🤢 This is absolutely awful. So even using `x:Bind`, as in #917, didn't worked? If that's the case, couldn't the fix be implemented directly in the `ReswPlus` package?

huynhsontung:
Never mind, I found a much better solution 85bc6d77c2423de9b9051350cf38eeeacc624ef7

> So even using x:Bind, as in https://github.com/huynhsontung/Screenbox/pull/917, didn't worked?

We can't use `x:Bind` everywhere, like in `ResourceDictionary`.

PR #927, opened 2026-06-06
refactor: GridViewItemTemplates to enable x:Bind

United600:
- Refactors the ResourceDictionary according to the [documentation](https://learn.microsoft.com/en-us/windows/apps/develop/data-binding/data-binding-in-depth#resource-dictionaries-with-xbind) to support `x:Bind` and ensure type safety for item templates
- Centralizes media items glyph logic in `GlyphConverter`
- Simplifies `CommonGridViewItem`
- Moves playlist ItemTemplate to `GridViewItemTemplates.xaml` dictionary

Since `IsPlaying` in `MediaViewModel` is a nullable boolean, it remains a Binding. Is it safe to change it?

https://github.com/huynhsontung/Screenbox/blob/599a474b49cf8cb07b79c0ab308191a382888af8/Screenbox.Core/ViewModels/MediaViewModel.cs#L96


Screenbox/Templates/GridViewItemTemplates.xaml
```diff
@@ -1,63 +1,89 @@
...
+    xmlns:strings="using:Screenbox.Strings"
+    xmlns:viewModels="using:Screenbox.Core.ViewModels">
 
-    <DataTemplate x:Key="AlbumGridViewItemTemplate">
+    <DataTemplate x:Key="AlbumGridViewItemTemplate" x:DataType="viewModels:AlbumViewModel">
         <controls:CommonGridViewItem
-            Title="{Binding Name, Converter={StaticResource AlbumTextConverter}, FallbackValue={strings:Resources Key=UnknownAlbum}}"
+            Title="{x:Bind converters:DefaultStringConverter.GetValueOrFallback(Name, strings:Resources.UnknownAlbum)}"
             Width="{StaticResource CommonGridViewItemWidth}"
-            Caption="{Binding ArtistName, Converter={StaticResource ArtistTextConverter}, FallbackValue={strings:Resources Key=UnknownAlbum}}"
```
United600, line 11: This fallback (`Caption`) was incorrect, right? I updated it to "UnknownArtist".

huynhsontung: Yup you are right

huynhsontung:
> Since IsPlaying in MediaViewModel is a nullable boolean, it remains a Binding. Is it safe to change it?

Probably best to create a converter to convert from `bool?` to `bool` for the controls if we want full `x:Bind` coverage.

United600:
> > Since IsPlaying in MediaViewModel is a nullable boolean, it remains a Binding. Is it safe to change it?
> 
> Probably best to create a converter to convert from `bool?` to `bool` for the controls if we want full `x:Bind` coverage.

I'm not sure I like that. Is there a technical reason why (and only) the `MediaViewModel.IsPlaying` is nullable? I think I'll change it to a boolean and see what breaks, then fix it from there.

huynhsontung:
It's a loosely state playback state translation. Null = not active, false = paused, true = playing.

Edit: typing that out I realized I could've just set not active and paused to false lol

United600:
Yeah, only a couple of hits that seem straightforward to resolve, and `false` can cover both states.

`CS0019	Operator '??' cannot be applied to operands of type 'bool' and 'bool'	Screenbox.Core\ViewModels\SeekBarViewModel.cs	234`

https://github.com/huynhsontung/Screenbox/blob/59714c15c06ecb5e790ee85b7d11d34c2c928410/Screenbox.Core/ViewModels/SeekBarViewModel.cs#L234

`CS0019	Operator '??' cannot be applied to operands of type 'bool' and 'bool'	Screenbox.Core\ViewModels\PlayQueuePanelViewModel.cs	110`

https://github.com/huynhsontung/Screenbox/blob/59714c15c06ecb5e790ee85b7d11d34c2c928410/Screenbox.Core/ViewModels/PlayQueuePanelViewModel.cs#L110

`CS0019	Operator '??' cannot be applied to operands of type 'bool' and 'bool'	Screenbox.Core\ViewModels\ArtistViewModel.cs	60`

https://github.com/huynhsontung/Screenbox/blob/59714c15c06ecb5e790ee85b7d11d34c2c928410/Screenbox.Core/ViewModels/ArtistViewModel.cs#L60

`CS0019	Operator '??' cannot be applied to operands of type 'bool' and 'bool'	Screenbox.Core\ViewModels\AlbumViewModel.cs	101`

https://github.com/huynhsontung/Screenbox/blob/59714c15c06ecb5e790ee85b7d11d34c2c928410/Screenbox.Core/ViewModels/AlbumViewModel.cs#L101

`CS0037	Cannot convert null to 'bool' because it is a non-nullable value type	Screenbox.Core\Coordinators\PlayQueueCoordinator.cs	525`

https://github.com/huynhsontung/Screenbox/blob/59714c15c06ecb5e790ee85b7d11d34c2c928410/Screenbox.Core/Coordinators/PlayQueueCoordinator.cs#L525

Screenbox/Converters/MediaGlyphConverter.cs
```diff
@@ -9,44 +7,13 @@ namespace Screenbox.Converters
...
             {
                 null => "\uea69",
-                IStorageItem item => Convert(item),
-                MediaViewModel { Source: StorageFile file } => Convert(file),
-                MediaViewModel { IsFromLibrary: true, MediaType: var mediaType } => Convert(mediaType),
+                IStorageItem item => GlyphConverter.ToStorageItemGlyph(item),
+                MediaViewModel viewModel => GlyphConverter.ToMediaGlyph(viewModel),
                 _ => "\ue774"
             };
         }
```
huynhsontung: Why don't we just create a static method in `GlyphConverter` and  `x:Bind` everything? Then we can remove this `MediaGlyphConverter`

United600: I attempted to completely remove the converter and did something similar to f932d0c (only a subset), but it resulted in garbage data. Probably missed something obvious.

huynhsontung: No worry, I think I did it. Seems like a natural next step.

huynhsontung (approved):

Screenbox/Converters/MediaGlyphConverter.cs
United600: I'm in the process of converting more converter uses to x:Bind functions.

United600:
Heads up, I'm seeing some unusual ordering on the `ArtistDetailsPage` with long lists after commit f932d0c.

https://github.com/user-attachments/assets/d1e1024e-3521-43aa-9f64-7c3509ae60e5

Screenbox/Converters/MediaGlyphConverter.cs
huynhsontung: Wait if it's not related to this let's merge this in first 😅

United600: It kinda was, but I'll hold off for now.

huynhsontung:
@United600 Can you try again to see if the problem goes away with the latest commit?

United600:
> @United600 Can you try again to see if the problem goes away with the latest commit?

Seems fine 👌.

United600:
Odd fix, but we'll take it. In the future, we should migrate this to a templated control.

Screenbox/Templates/GridViewItemTemplates.xaml
```diff
@@ -1,63 +1,89 @@
...
                 <FontIconSource FontSize="{StaticResource GridItemPlaceholderIconFontSize}" Glyph="&#xE77B;" />
             </controls:CommonGridViewItem.PlaceholderIconSource>
         </controls:CommonGridViewItem>
     </DataTemplate>
 
-    <DataTemplate x:Key="MediaGridViewItemTemplate">
+    <DataTemplate x:Key="MediaGridViewItemTemplate" x:DataType="viewModels:MediaViewModel">
         <controls:CommonGridViewItem
-            Title="{Binding Name}"
+            Title="{x:Bind Name}"
```
United600:
```suggestion
            Title="{x:Bind Name, Mode=OneWay}"
```

I think we may need to make the `Title` and `Caption` binding mode "OneWay", otherwise this issue sometimes appears at launch. I think the binding mode isn't the issue, the media detection at launch is but until then...

https://github.com/user-attachments/assets/4a37d3ba-126f-4812-a578-e1c02094f178

Screenbox/Templates/GridViewItemTemplates.xaml
```diff
@@ -1,63 +1,89 @@
...
             Width="{StaticResource CommonGridViewItemWidth}"
-            Caption="{Binding Caption}"
-            IsPlaying="{Binding IsPlaying}"
+            AutomationProperties.Name="{x:Bind helpers:ItemLabelHelper.GetAccessibleNameForItem(Name, Caption)}"
+            Caption="{x:Bind Caption}"
```
United600:
```suggestion
            Caption="{x:Bind Caption, Mode=OneWay}"
```

PR #930, opened 2026-06-10
feat: add drag-and-drop and reordering support to playlist details

United600:
Adds drag-and-drop for reordering playlist items and adding items from the explorer, and adds a no-content visual state for empty playlists.
Saving playlist files now uses `SerializeToUtf8Bytes` for improved performance as recommended in the [documentation](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/how-to#serialize-to-utf-8).
Ensures the playlist export file is written using UTF‑8 encoding.


Screenbox.Core/ViewModels/PlaylistViewModel.cs
```diff
@@ -91,7 +102,28 @@ public async Task AddItemsAsync(IReadOnlyList<MediaViewModel> items)
             Items.Add(Items.Contains(item) ? new MediaViewModel(item) : item);
         }
 
-        await SaveAsync();
+        //await SaveAsync();
```
United600, line 108: Commented out to prevent duplicate calls, as adding to a collection is now being tracked via the `Items_CollectionChanged`. Should have removed it, but wasn't certain.

Screenbox/Pages/PlaylistDetailsPage.xaml.cs
```diff
@@ -85,6 +88,33 @@ private void PlaylistDetailsPage_OnLoaded(object sender, RoutedEventArgs e)
...
+    private void ItemList_OnDragOver(object sender, DragEventArgs e)
+    {
+        e.Handled = true;
+        e.AcceptedOperation = e.DataView.Contains(StandardDataFormats.StorageItems)
+            ? DataPackageOperation.Copy
+            : DataPackageOperation.None;
+
+        if (e.DragUIOverride != null)
+        {
+            e.DragUIOverride.Caption = Strings.Resources.AddToPlaylist;
```
United600, line 100: This, and the `PlayQueueControl.xaml.cs` one, override are way too finicky, the one in `MainPage.xaml.cs` takes precedence most of the time.

Screenbox.Core/ViewModels/PlaylistViewModel.cs
```diff
@@ -151,4 +183,22 @@ private MediaViewModel ToMediaViewModel(PersistentMediaRecord record)
...
+                {
+                    await SaveAsync();
+                }
+                catch (Exception ex)
+                {
+                    LogService.Log(ex);
+                }
+            });
+        }
+    }
```
United600: This was the quickest way to enable it without issues, but it could be improved by adding a debounce timer to batch changes. I tried to avoid using the `CommunityToolkit.WinUI` namespace.

Screenbox.Core/Services/FilesService.cs
```diff
@@ -132,8 +132,9 @@ public async Task SaveToDiskAsync<T>(StorageFile file, T source)
     {
         if (file.Name.EndsWith(".json", StringComparison.OrdinalIgnoreCase))
         {
-            var json = JsonSerializer.Serialize(source);
-            await FileIO.WriteTextAsync(file, json);
+            var options = new JsonSerializerOptions { WriteIndented = true };
```
United600: Perhaps redundant, but it helped with debugging. Though it causes a size regression, how much does readability matter?

Screenbox.Core/ViewModels/PlaylistViewModel.cs
huynhsontung: I don't think debouncer is an issue. It's pretty small. Even if we don't use `CommunityToolkit` for any reason, we can copy the debouncer logic over. We can use debounce here. Without it, the playlist is written for every item change, which can be a lot.

Screenbox.Core/ViewModels/PlaylistViewModel.cs
```diff
@@ -151,4 +183,22 @@ private MediaViewModel ToMediaViewModel(PersistentMediaRecord record)
 
         return media;
     }
+
+    private async Task UpdatePlaylistAsync()
```
huynhsontung: If the save task is called through the dispatcher queue then there is no reason for this method to be an async Task

Screenbox.Core/ViewModels/PlaylistViewModel.cs
```diff
@@ -33,18 +35,27 @@ public partial class PlaylistViewModel : ObservableRecipient
...
     }
 
-    private void Items_CollectionChanged(object sender, System.Collections.Specialized.NotifyCollectionChangedEventArgs e)
+    private void Items_CollectionChanged(object sender, NotifyCollectionChangedEventArgs e)
     {
         OnPropertyChanged(nameof(ItemsCount));
+
+        // Ensures the playlist is up to date when items are reordered, as the ListViewBase
+        // reorder operation sends Remove/Add actions instead of a Move action.
+        if (e.Action == NotifyCollectionChangedAction.Add)
```
huynhsontung, line 58: Why only filter for `NotifyCollectionChangedAction.Add` here? What about the remove and move operations?

Screenbox.Core/ViewModels/PlaylistViewModel.cs
United600: Agreed. I tried dropping a folder with hundreds of items and it worked fine, but there's still no point in doing extra work.

Screenbox.Core/ViewModels/PlaylistViewModel.cs
United600: `Remove` is already handled by the Remove command, and I don't think `Move` is used anywhere in this collection

huynhsontung: Can't you reorder the playlist?

Screenbox.Core/Services/FilesService.cs
huynhsontung: Yea readability is not that important in this case.

Screenbox.Core/ViewModels/PlaylistViewModel.cs
United600: Yes, but as mentioned in the comment:

```csharp
// Ensures the playlist is up to date when items are reordered, as the ListViewBase
// reorder operation sends Remove/Add actions instead of a Move action.
```

PR #934, opened 2026-06-15
fix: play queue move up/down command CanExecute status

United600:
Fixes a regression introduced in #870. Evaluating only on `IsAllSelected` changes isn't sufficient.

**Before**

https://github.com/user-attachments/assets/5ee4843f-601b-4143-942a-d476d88ff731

**After**

https://github.com/user-attachments/assets/02e1fe26-2381-422d-9657-06f93f67ed5e

### DIFF
--- a/Screenbox.Core/ViewModels/PlayQueuePanelViewModel.cs
+++ b/Screenbox.Core/ViewModels/PlayQueuePanelViewModel.cs
@@ -1,9 +1,8 @@
-﻿#nullable enable
+#nullable enable
 
 using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
-using System.ComponentModel;
 using System.Linq;
 using System.Threading.Tasks;
 using CommunityToolkit.Mvvm.ComponentModel;
@@ -60,7 +59,7 @@ public sealed partial class PlayQueuePanelViewModel : ObservableRecipient
         Queue.Items.CollectionChanged += ItemsOnCollectionChanged;
 
         Selection.SetItemsSource(Queue.Items);
-        Selection.PropertyChanged += Selection_OnPropertyChanged;
+        Selection.SelectedItemsChanged += Selection_SelectedItemsChanged;
     }
 
     private void ItemsOnCollectionChanged(object sender, NotifyCollectionChangedEventArgs e)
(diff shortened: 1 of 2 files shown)


Screenbox.Core/ViewModels/PlayQueuePanelViewModel.cs
```diff
@@ -60,7 +59,7 @@ public PlayQueuePanelViewModel(
         Queue.Items.CollectionChanged += ItemsOnCollectionChanged;
 
         Selection.SetItemsSource(Queue.Items);
-        Selection.PropertyChanged += Selection_OnPropertyChanged;
+        Selection.SelectedItemsChanged += Selection_SelectedItemsChanged;
```
United600: I thought about subscribing directly to the `Selection.SelectedItems.SelectionChanged` event, but I wasn't keen on the idea of a "double" subscription.

Screenbox.Core/ViewModels/PlayQueuePanelViewModel.cs
```diff
@@ -204,13 +200,13 @@ public Task EnqueueDroppedItemsAsync(IReadOnlyList<IStorageItem> items, int inse
 
     private bool HasSelection() => Selection.SelectedItems.Count > 0;
 
-    private bool IsSelectedItemNotFirst(IList<object>? selectedItems) =>
```
United600, line 207: For some reason it only worked properly when the buttons were in the overflow menu, so I had to remove the parameter and use `Selection.SelectedItems` directly.

Screenbox.Core/ViewModels/SelectionViewModel.cs
```diff
@@ -50,6 +50,9 @@ public sealed partial class SelectionViewModel : ObservableObject
 
     private IReadOnlyCollection<object>? _sourceCollection;
 
+    /// <inheritdoc cref="ObservableCollection{T}.CollectionChanged"/>
+    public event NotifyCollectionChangedEventHandler? SelectedItemsChanged;
```
United600:
```suggestion
    public event Action? SelectedItemsChanged;
```

What do you think? For starters, the name is a bit misleading.
There is probably a simpler way to do it, but I didn't want to add another `ObservableProperty` just for tracking the count change.

Screenbox.Core/ViewModels/SelectionViewModel.cs
```diff
@@ -106,6 +109,7 @@ private void ClearSelection()
 
     private void SelectedItems_OnCollectionChanged(object sender, NotifyCollectionChangedEventArgs e)
     {
+        SelectedItemsChanged?.Invoke(sender, e);
```
United600:
```suggestion
        SelectedItemsChanged?.Invoke();
```

Continuation of the above suggestion.

Screenbox.Core/ViewModels/PlayQueuePanelViewModel.cs
huynhsontung: Since `SelectedItems` is a get-only property, I don't have any concern if you directly subscribe to `SelectedItems.SelectionChanged`.

Screenbox.Core/ViewModels/SelectionViewModel.cs
huynhsontung: I think having a custom event here is more confusing. I don't think we use this pattern anywhere else in the Core project. We usually just subscribe directly to the ObservableCollection since they are usually get-only properties.

Screenbox.Core/ViewModels/PlayQueuePanelViewModel.cs
United600: Oh! It felt weird, but it's good to know.

huynhsontung (approved):

PR #935, opened 2026-06-15
fix: F11 keyboard shortcut

### LINKED ISSUE #933: [Feature Request] F11 shortcut for Fullscreen
It's not a problem, but it's a nice to have feature. It's a quality of life feature, because this shortcut used in alot of other application or websites.

Add a new shortcut to fullscreen the video, default to f11.

There's no alternative i found, i've always used the fullscreen button to go into the fullscreen mode

**Additional context**
That's all. Thanks for seeing my issue, hope it'll added soon to the apps. I love this app, thanks for creating such an app ^^

United600:
Closes #933. The accelerator was only present in the `MenuFlyoutItem`, so it was only functional when resizing the app until the button was hidden and the menu item became active, after that it would work for the remaining of the app lifetime.

I wonder if we should remove both accelerators and implement them the same way we do in the `PlayerPage`, fewer moving parts.

### DIFF
--- a/Screenbox/Controls/PlayerControls.xaml
+++ b/Screenbox/Controls/PlayerControls.xaml
@@ -1,4 +1,4 @@
-﻿<UserControl
+<UserControl
     x:Class="Screenbox.Controls.PlayerControls"
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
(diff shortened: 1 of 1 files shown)


United600:
We should also add a guard to prevent triggering when the play queue is empty. Currently, it still triggers after clearing the queue.

huynhsontung:
I like the idea of handling it on the `PlayerPage`

United600:
> I like the idea of handling it on the `PlayerPage`

It sounded better in theory, we'd still need to keep the commands for the buttons unless we refactor it further and move them to the `CommonViewModel`.

PR #936, opened 2026-06-15
fix: play queue add folder command

United600:
Fixes a regression introduced in #701, where the current queue was being replaced instead of appended to.

### DIFF
--- a/Screenbox.Core/ViewModels/PlayQueuePageViewModel.cs
+++ b/Screenbox.Core/ViewModels/PlayQueuePageViewModel.cs
@@ -1,4 +1,4 @@
-﻿#nullable enable
+#nullable enable
 
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


United600:
BTW, I found the responsibility split between `PlayQueuePanelViewModel` and `PlayQueuePageViewModel` a bit particular. Why is the AddFiles command in one, while AddUrl and AddFolder are in the other?

huynhsontung:
Because the `PlayQueuePanelViewModel` is used in the playlist flyout on the `PlayerPage,` and that flyout only has a button to add files, not folders or URLs. The add folder and URL functions are only available from the `PlayQueuePage`. Is this ideal? Idk. That's a different discussion.

United600:
> Because the `PlayQueuePanelViewModel` is used in the playlist flyout on the `PlayerPage,` and that flyout only has a button to add files, not folders or URLs. The add folder and URL functions are only available from the `PlayQueuePage`. Is this ideal? Idk. That's a different discussion.

Ah! That's true, I completely forgot about the button in the flyout.

Issue #947, opened 2026-06-29
Improve video player with color inversion mode and chapter list


nvsirkhvn404:
## Is your feature request related to a problem? Please describe.

I have two feature requests that would improve the experience of watching long lecture videos.

### 1. Video color inversion

Many lecture videos contain white presentation slides that are very bright, especially during long study sessions or at night. This can cause eye strain.

### 2. Chapter list

When a video contains chapters, only the chapter markers and names are visible on the timeline. It's difficult to see all chapters at once or quickly navigate to a specific one.

### 1. Video color inversion

Add an optional toggle that inverts only the video colors (not the entire UI). This would make white slides appear dark and reduce eye strain.

### 2. Chapter list

Add a scrollable chapter list similar to YouTube that:

* Shows all chapter titles.
* Displays timestamps.
* Lets users jump directly to any chapter.
* Highlights the currently playing chapter.
* Shows the total number of chapters.

## Describe alternatives you've considered

For color inversion, browser extensions or system-wide color inversion affect the entire page instead of just the video.

For chapter navigation, the timeline markers help but become difficult to use when a video has many chapters.

## Additional context

Both features would significantly improve the experience of watching long educational videos and lectures.

> **Note:** This issue was written with the help of AI for clarity and formatting. The feature requests themselves are my own.

PR #951, opened 2026-06-30
feat: add chapter button with selectable chapter list

### LINKED ISSUE #217: Add ability to jump to a chapter

United600:
Closes #217, #508, and partially #947. Adds a chapter menu button to the playback controls, allowing seamless browsing and chapter navigation.

I left the `TimeDisplay` control untouched, though we should strip out the chapter and title bits. I also didn't add a dedicated view model for the `ChapterPickerControl`, opting instead for a simpler implementation.

PS. Not sure what causes the leading space in `ChapterCue` names, but it appeared in all my test files, so I added a `TrimStart()` fix.

https://github.com/user-attachments/assets/735c7683-84c1-404f-a762-001765480b32

### DIFF
--- a/Screenbox.Core/ViewModels/SeekBarViewModel.cs
+++ b/Screenbox.Core/ViewModels/SeekBarViewModel.cs
@@ -229,12 +229,14 @@ public sealed partial class SeekBarViewModel :
         UpdateProgress(newPosition);
     }
 
+    /// <summary>
+    /// Seeks the media playback position to the specified chapter.
+    /// </summary>
+    /// <param name="chapter">The chapter to seek.</param>
     [RelayCommand]
-    private void SeekToChapter(ChapterCue? chapter)
+    private void SeekToChapter(ChapterCue chapter)
     {
-        if (chapter is null) return;
-
-        UpdatePosition(chapter.StartTime, isOffset: false, debounce: false);
+        Messenger.Send(new ChangeTimeRequestMessage(chapter.StartTime, debounce: false));
     }
 
     private void RestoreLastPosition(MediaViewModel media)
--- a/Screenbox.Core/ViewModels/SeekBarViewModel.cs
+++ b/Screenbox.Core/ViewModels/SeekBarViewModel.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Collections.ObjectModel;
 using CommunityToolkit.Mvvm.ComponentModel;
+using CommunityToolkit.Mvvm.Input;
 using CommunityToolkit.Mvvm.Messaging;
 using CommunityToolkit.Mvvm.Messaging.Messages;
 using CommunityToolkit.WinUI;
@@ -45,6 +46,8 @@ public sealed partial class SeekBarViewModel :
     [ObservableProperty] private bool _shouldShowPreview;
 
     [ObservableProperty] private bool _shouldHandleKeyDown;
+
+    [ObservableProperty] private ChapterCue? _currentChapterCue;
 
     public ObservableCollection<ChapterCue> Chapters { get; }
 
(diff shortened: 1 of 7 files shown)


Screenbox/Controls/PlayerControls.xaml
```diff
@@ -443,12 +444,51 @@
...
+                            </Style>
+                        </Flyout.FlyoutPresenterStyle>
+                        <controls:ChapterPickerControl
+                            ChapterCommand="{x:Bind SeekBar.ViewModel.SeekToChapterCommand}"
+                            ChaptersSource="{x:Bind SeekBar.ViewModel.Chapters}"
+                            SelectedChapter="{x:Bind SeekBar.ViewModel.CurrentChapterCue, Mode=TwoWay}" />
+                    </Flyout>
+                </Button.Flyout>
+                <StackPanel Orientation="Horizontal">
+                    <TextBlock Text="{x:Bind SeekBar.ViewModel.CurrentChapterCue.Title.TrimStart(), Mode=OneWay}" Typography.NumeralAlignment="Tabular" />
```
huynhsontung, line 480: I didn't know you could bind to a function at the end like this.

Screenbox.Core/ViewModels/SeekBarViewModel.cs
huynhsontung, line 1: So SeekBar doesn't really use the chapter information. We are just reusing the view model. Shouldn't it make more sense to put all the chapter-related code in a `ChapterPickerViewModel`?

Screenbox/Controls/PlayerControls.xaml
```diff
@@ -443,12 +444,51 @@
...
+                Text="•"
+                Visibility="{x:Bind GetChapterVisibility(ViewModel.PlayerShowChapters, SeekBar.ViewModel.Chapters.Count), Mode=OneWay}" />
+            <Button
+                x:Name="ChapterButton"
+                Width="Auto"
+                Padding="4,0,4,1"
+                AutomationProperties.Name="{x:Bind SeekBar.ViewModel.CurrentChapterCue.Title.TrimStart(), Mode=OneWay}"
+                IsEnabled="{x:Bind IsEnabled, Mode=OneWay}"
+                Style="{StaticResource PlayerButtonStyle}"
+                Visibility="{x:Bind GetChapterVisibility(ViewModel.PlayerShowChapters, SeekBar.ViewModel.Chapters.Count), Mode=OneWay}">
```
huynhsontung, line 462: I think `ChapterPickerControl` should wrap the whole button, rather than just the flyout.

Screenbox/Controls/PlayerControls.xaml
United600: Super handy, but some of them just don't get suggested for whatever reason.

Screenbox.Core/ViewModels/SeekBarViewModel.cs
United600: It does, no?

It works similarly to the `ChapterProgressBar`. We could move the chapter logic, but a dedicated `ChapterPickerViewModel` seems unnecessary and would likely add repetition rather than reduce complexity.

Maybe consolidating all chapter-related logic within the `ChapterViewModel`? Still not entirely convinced.

Screenbox/Controls/PlayerControls.xaml
United600: Not sure I like that. Would the same logic apply to the `AudioAndCaptionButton`? When I first started, I even thought about putting everything directly in the `PlayerControls.xaml` 😅.

huynhsontung: After some discussion on Discord, it's probably better to have it as is.

Screenbox.Core/ViewModels/SeekBarViewModel.cs
huynhsontung: Yea. I take that back. After revisiting how `TimeDisplay` works, I don't mind this anymore.

PR #958, opened 2026-07-04
chore: update third-party notice

United600:


huynhsontung:
We are not using EF Core btw. Just the Microsoft.Data.Sqlite. I guess it's still part of EF Core.

United600:
> We are not using EF Core btw. Just the Microsoft.Data.Sqlite. I guess it's still part of EF Core.

Right. So, should we add a link to the repository from the NuGet package? https://github.com/dotnet/dotnet

And the name?

huynhsontung:
We should follow whatever information is posted on the NuGet package

https://www.nuget.org/packages/Microsoft.Data.Sqlite/10.0.9#readme-body-tab

PR #959, opened 2026-07-05
fix: AOT compiler warnings

huynhsontung:


United600:
CsWinRT3 should be nearing release, and it should line up well with our migration timeline, meaning all those "partial" will end up being unnecessary.

From the [spec](https://github.com/microsoft/CsWinRT/blob/staging/3.0/docs/cswinrt3.0-spec.md): "- No need to manually make every single type partial anymore"

huynhsontung:
I couldn't get CsWinRT3 preview to work. I think it's fine to make these changes. We don't know when CsWinRT3 will be released.

United600:
> I couldn't get CsWinRT3 preview to work. I think it's fine to make these changes. We don't know when CsWinRT3 will be released.

That's a shame. Does this imply we can already build using CsWinRT2?

PR #961, opened 2026-07-06
refactor: SeekBar pointer wheel event

United600:


United600:
🤔 Surely the fixed mouse wheel step applies here as well. I'll have to test it later.

huynhsontung (approved):

PR #973, opened 2026-07-10
refactor: migrate most System.Text.Json usage to source generation

United600:
Adopted source generation per the [document](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/source-generation) (and [this](https://devblogs.microsoft.com/dotnet/try-the-new-system-text-json-source-generator/)). Most of the serialization paths were updated,  the rest couldn't be ported and might only be possible with extensive refactoring.

Not sure if this [section](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/source-generation#disable-reflection-defaults) applies to .NET Native, it's unclear whether pre‑SDK projects support that property setting.

Also fixed many of the nullable reference warnings in `Screenbox.Lively`, but some still remain. The rest can be tackled in a future PR.


Screenbox.Core/Models/Serialization/CoreJsonContext.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+using System.Text.Json.Serialization;
+
+namespace Screenbox.Core.Models.Serialization;
+
+[Obsolete("Remove the class once the transition from Protobuf to Microsoft.Data.Sqlite has been in place for some time.")]
```
huynhsontung: I think this legacy playlist conversion flow is here to stay, or maybe till after we release v1.0. We don't want an app update to remove user playlists.

huynhsontung: In that case, we probably should reword the message into something like "This class will be removed in app version 1.0"

PR #976, opened 2026-07-12
chore: bump nuget packages and resolve SQLite vulnerabilities

huynhsontung:
Also pin the `SQLitePCLRaw.lib.e_sqlite3` package to 2.1.12 to avoid SQLite vulnerabilities.

### DIFF
--- a/Screenbox.Core/Screenbox.Core.csproj
+++ b/Screenbox.Core/Screenbox.Core.csproj
@@ -329,7 +329,7 @@
       <Version>5.0.7</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions">
-      <Version>10.0.9</Version>
+      <Version>10.0.10</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.NETCore.UniversalWindowsPlatform">
       <Version>6.2.14</Version>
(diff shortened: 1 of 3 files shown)


United600:
Can we even use that package? Based on [this](https://github.com/dotnet/efcore/issues/38257#issuecomment-4928826023), it certainly seems like the answer is no.

And if we're going to pin a transitive package, why not just use [CPM](https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management) instead?

huynhsontung:
> Can we even use that package? Based on [this](https://github.com/dotnet/efcore/issues/38257#issuecomment-4928826023), it certainly seems like the answer is no.

It still works fine when I tested on my machine. I also pinned the SQLite lib to 2.1.12 instead. The current deprecated one is 2.1.11 so hopefully it's just unsupported on paper and not actual ABI changes.

> And if we're going to pin a transitive package, why not just use [CPM](https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management) instead?

I don't see the benefits actually.

United600:
> I don't see the benefits actually.

Sure, but it would save us from having to manually update dependencies across multiple projects (like the `Microsoft.Extensions.DependencyInjection`).

United600 (approved):

huynhsontung:
> Sure, but it would save us from having to manually update dependencies across multiple projects

When we move to AOT there is a nuget management for the whole solution. It updates and all nuget across all projects at the same time.

PR #978, opened 2026-07-13
fix: ensure accessibility of files from the Most Recently Used list before opening

huynhsontung:


huynhsontung:
This is probably unwanted after some more considerations

PR #979, opened 2026-07-13
style: add tool tips to trimmed chapter names

United600:
Continuation of #951. 

- Adds tooltips to chapter titles when they’re truncated in the picker list or in the controls
- Streamlines the `TimeDisplayControl`, since it no longer manages any chapter‑related logic
- Adds persistence setting for the time display preference, allowing users to keep their choice between remaining and elapsed time
- Unwraps the Previous/Next buttons from the panel that enforced a fixed ordering, shifting their ordering logic into a visual state to correctly handle RTL layouts

### DIFF
--- a/Screenbox.Core/Services/SettingsService.cs
+++ b/Screenbox.Core/Services/SettingsService.cs
@@ -1,4 +1,4 @@
-﻿#nullable enable
+#nullable enable
 
 using System;
 using System.Linq;
@@ -32,6 +32,7 @@ public sealed class SettingsService : ISettingsService
     private const string MaxVolumeKey = "Values/MaxVolume";
     private const string PersistentRepeatModeKey = "Values/RepeatMode";
     private const string PersistentSubtitleLanguageKey = "Values/SubtitleLanguage";
+    private const string PlayerShowRemainingTimeKey = "Player/ShowRemainingTime";
     private const string PlayerShowChaptersKey = "Player/ShowChapters";
     private const string PrivacyPersistPlaybackPosition = "Privacy/PersistPlaybackPosition";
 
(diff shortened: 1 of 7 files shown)


huynhsontung:
So we switched from letting users switch between elapsed and remaining time by clicking the TimeDisplay to offering it as a Settings option? Why is this change?

If the feature is about persistence, should we persist the TimeDisplay state directly and skip the settings?

United600:
It still works by tapping the control, but now the preference is saved between sessions. I didn't add a new UI setting.

huynhsontung:
Oh I read it wrong then. My bad.

United600:
TBF, the wording leaves things a bit open to interpretation.

Screenbox.Core/Services/SettingsService.cs
```diff
@@ -32,6 +32,7 @@ public sealed class SettingsService : ISettingsService
     private const string MaxVolumeKey = "Values/MaxVolume";
     private const string PersistentRepeatModeKey = "Values/RepeatMode";
     private const string PersistentSubtitleLanguageKey = "Values/SubtitleLanguage";
+    private const string PlayerShowRemainingTimeKey = "Player/ShowRemainingTime";
```
huynhsontung: It's not documented anywhere but if it's a key for state storing only and not an actual user facing setting, e.g. persistent volume, start the key with `Values/` pls. And call the property Persistent*, like `PersistentShowRemainingTime`

United600: Ideally this would be named `PersistentPlayerTimeDisplayMode` or something similar, but the setting isn't significant enough to justify creating a dedicated enum.

huynhsontung (approved):

PR #982, opened 2026-07-16
feat: add multi-selection support to Playlists page

United600:
- Adds multi‑item selection functionality to `PlaylistsPage`
- Updates the PlaylistsPage layout to better integrate the selection bar
- Improves the minimal visual state shown when `PlaylistsPage` content is unavailable
- Updates `SelectionBarControl` and `PlayQueueControl.ActionButtonPanel` iconography

#### Multi-Selection

https://github.com/user-attachments/assets/dbe908e4-e357-461e-bdb6-7f236545f935

#### Layout

Screenbox new playlists page layout

#### VisualState

| Before | After |
|-|-|
|Screenbox playlists page NoContentVisualState|Screenbox new playlists page NoContent visual state|


huynhsontung:
Not sure I'm a fan of the new Import button position. The option to add something to the page has always been in the top right corner. This is true for the Home page, Music page, Videos page, and Play queue page. The new layout feels random. The import button location feels like it should be more consistent if there is a sorting option there.

United600:
> Not sure I'm a fan of the new Import button position. The option to add something to the page has always been in the top right corner. This is true for the Home page, Music page, Videos page, and Play queue page. The new layout feels random. The import button location feels like it should be more consistent if there is a sorting option there.

Yeah, I get what you mean. We definitely need to add something to the sub-header to support the selection command bar. If we had a play button, I agree the add button would make more sense in the top‑right, but alas. How about simply moving the import button next to the new button instead? Or moving one of them to the top-right? I'm fine with anything as long as it doesn't involve adding a split button to the sub‑header.

huynhsontung:
I'm good with moving the import button next to the new button in the sub-header.

United600:
> I'm good with moving the import button next to the new button in the sub-header.

It'll likely flow better and keeps the right side open for a sort button.

PR #987, opened 2026-07-20
refactor: item flyouts bindings

United600:


Screenbox/Behaviors/ListViewContextTriggerBehavior.cs
```diff
@@ -1,101 +1,128 @@
...
+            Flyout?.ShowAt(element);
         }
+
+        e.Handled = true;
+    }
+
+    private static bool ShouldShowFlyout(object? context)
+    {
+        return context is not StorageItemViewModel { Media: null };
     }
```
United600: TBH, I'm not a fan of this method. This behavior shouldn't know anything about the view model, so we should switch back to the event.

Screenbox/Pages/AlbumsPage.xaml.cs
```diff
@@ -37,7 +39,7 @@ public AlbumsPage()
 
     private void ViewModelOnPropertyChanged(object? sender, PropertyChangedEventArgs e)
     {
-        if (e.PropertyName == nameof(SongsPageViewModel.SortBy))
+        if (e.PropertyName == nameof(AlbumsPageViewModel.SortBy))
```
United600, line 42: That wasn't correct, right?

Screenbox/Behaviors/ListViewContextTriggerBehavior.cs
huynhsontung: Agree. I also think using the `ContextRequested` event makes more sense here. The parent control can intercept the flyout and decide whether it should be shown.

Screenbox/Pages/AlbumsPage.xaml.cs
huynhsontung: Yea. That was a copy-and-paste mistake. Nice catch. Lucky that it compiles to the same thing.

Screenbox/Behaviors/ListViewContextRequestedEventArgs.cs
```diff
@@ -1,17 +1,16 @@
...
+
+public class ListViewContextRequestedEventArgs : EventArgs
 {
-    public class ListViewContextRequestedEventArgs : EventArgs
-    {
-        public SelectorItem Item { get; }
+    public SelectorItem Item { get; }
 
-        public bool Handled { get; set; }
+    public bool ShouldShowFlyout { get; set; } = true;
```
United600: I would also reinstate the `Handled` property to maintain consistency with WinRT events.

huynhsontung (approved):

PR #989, opened 2026-07-22
refactor: PropertyView bindings

United600:
Restores `PropertyViewModel` to the **Core** project, partially reverting the modifications from #794.

### DIFF
--- a/Screenbox.Core/Enums/Property.cs
+++ b/Screenbox.Core/Enums/Property.cs
@@ -1,0 +1,23 @@
+namespace Screenbox.Core.Enums;
+
+public enum Property
+{
+    Title,
+    Subtitle,
+    Year,
+    Producers,
+    Writers,
+    Length,
+    Resolution,
+    BitRate,
+    ContributingArtists,
+    Album,
+    AlbumArtist,
+    Composers,
+    Genre,
+    Track,
+    FileType,
+    ContentType,
+    Size,
+    LastModified,
+}
--- a/Screenbox/Controls/PropertiesView.xaml
+++ b/Screenbox/Controls/PropertiesView.xaml
@@ -32,7 +32,7 @@
                         Margin="0,0,8,0"
                         Foreground="{ThemeResource TextFillColorSecondaryBrush}"
                         Style="{StaticResource CaptionTextBlockStyle}"
-                        Text="{x:Bind converters:ResourceNameToResourceStringConverter.FromName(Name)}"
+                        Text="{x:Bind converters:ResourceNameToResourceStringConverter.FromEnum(Key)}"
                         TextTrimming="CharacterEllipsis" />
 
                     <TextBlock
--- a/Screenbox/Controls/PropertiesView.xaml
+++ b/Screenbox/Controls/PropertiesView.xaml
@@ -5,6 +5,7 @@
     xmlns:converters="using:Screenbox.Converters"
     xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:models="using:Screenbox.Core.Models"
     xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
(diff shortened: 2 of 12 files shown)


Screenbox.Core/Enums/Property.cs
```diff
@@ -0,0 +1,23 @@
+namespace Screenbox.Core.Enums;
+
+public enum Property
```
United600: I'm not loving this name.

Screenbox/Controls/PropertiesView.xaml
```diff
@@ -1,58 +1,66 @@
...
     xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
     xmlns:strings="using:Screenbox.Strings"
     xmlns:triggers="using:CommunityToolkit.WinUI"
     d:DesignHeight="300"
     d:DesignWidth="400"
     mc:Ignorable="d">
     <UserControl.Resources>
-        <DataTemplate x:Key="PropertyItemTemplate">
+        <DataTemplate x:Key="PropertyItemTemplate" x:DataType="models:MediaMetadata">
             <UserControl>
```
United600, line 18: Also less than ideal, unfortunate that visual states force this `UserControl` wrap.

Screenbox.Core/Enums/Property.cs
huynhsontung: Me neither 🤔

huynhsontung: Maybe `MediaProperty`?

United600: > Maybe `MediaProperty`?

Definitely much better (we ignore the file‑related properties 😉), as long as you're fine replacing the `ResourceNameToResourceString` converter with a switch statement. I am.

huynhsontung: Ohh I see. Yea if you want to rename the property, just use a switch statement. We probably won't change the resource keys in Crowdin so we don't have to worry about the switch statement go out of sync with Crowdin.

Actually the switch statement would enforce compile time enum mapping to see if it handles all the values, so it can never go out of sync. I'm sold!

Screenbox/Helpers/ItemLabelHelper.cs
```diff
@@ -112,6 +112,37 @@ public static string FormatAccessibleSearchSuggestionName(SearchSuggestionType t
...
+            MediaProperty.Album => Strings.Resources.PropertyAlbum,
+            MediaProperty.AlbumArtist => Strings.Resources.PropertyAlbumArtist,
+            MediaProperty.Composers => Strings.Resources.PropertyComposers,
+            MediaProperty.Genre => Strings.Resources.PropertyGenre,
+            MediaProperty.Track => Strings.Resources.PropertyTrack,
+            MediaProperty.FileType => Strings.Resources.PropertyFileType,
+            MediaProperty.ContentType => Strings.Resources.PropertyContentType,
+            MediaProperty.Size => Strings.Resources.PropertySize,
+            MediaProperty.LastModified => Strings.Resources.PropertyLastModified,
+            _ => property.ToString()
```
huynhsontung, line 142: Remove this line. The compiler statically checks for all enum cases and throws you a warning if you don't handle them all.

huynhsontung (approved):

Screenbox/Helpers/ItemLabelHelper.cs
United600: But won't that trigger a new warning? Handle default case.

huynhsontung: If all enum cases are handled then it shouldn't trigger a warning, I think

United600: Usually yes, but with switch expressions the default case is "required".

https://github.com/user-attachments/assets/56272556-1776-45f0-8b28-0b1270235346

huynhsontung: Hmm maybe I misremembered. It's okay as-is

PR #991, opened 2026-07-22
feat: add open artist command to albums page item flyout

United600:
https://github.com/user-attachments/assets/cc1d40a1-a374-4397-81e9-9b399a113155

### DIFF
--- a/Screenbox/Pages/AlbumsPage.xaml
+++ b/Screenbox/Pages/AlbumsPage.xaml
@@ -54,6 +54,11 @@
                     CommandParameter="{x:Bind ViewModel.ContextAlbum, Mode=OneWay}"
                     Icon="{ui:FontIcon Glyph=&#xE93C;}"
                     Text="{strings:Resources Key=ShowAlbum}" />
+                <MenuFlyoutItem
+                    Command="{x:Bind Common.OpenArtistCommand}"
+                    CommandParameter="{x:Bind ViewModel.ContextAlbum.RelatedSongs[0].MainArtist, FallbackValue={x:Null}, Mode=OneWay}"
+                    Icon="{ui:FontIcon Glyph=&#xE77B;}"
+                    Text="{x:Bind strings:Resources.ShowArtist}" />
             </MenuFlyout>
         </ResourceDictionary>
     </Page.Resources>


United600:
Before the item flyout bindings refactor, I had a branch that did something like this:

`AlbumViewModel.cs`

```csharp
public sealed partial class AlbumViewModel : ObservableRecipient
{
    ...
    public ArtistViewModel? MainArtist => RelatedSongs.Count > 0 ? RelatedSongs[0].MainArtist : null;
    ...
}
```

`AlbumsPage.xaml`

```xml
<MenuFlyout x:Key="AlbumFlyout">
    ....
    <MenuFlyoutItem
        Command="{x:Bind Common.OpenArtistCommand}"
        CommandParameter="{Binding MainArtist}"
        Icon="{ui:FontIcon Glyph=&#xE77B;}"
        Text="{x:Bind strings:Resources.ShowArtist}" />
</MenuFlyout>
```

If you prefer that approach, we can tweak it for the new bindings setup.

Screenbox/Pages/AlbumsPage.xaml
```diff
@@ -54,6 +54,11 @@
                     CommandParameter="{x:Bind ViewModel.ContextAlbum, Mode=OneWay}"
                     Icon="{ui:FontIcon Glyph=&#xE93C;}"
                     Text="{strings:Resources Key=ShowAlbum}" />
+                <MenuFlyoutItem
+                    Command="{x:Bind Common.OpenArtistCommand}"
+                    CommandParameter="{x:Bind ViewModel.ContextAlbum.RelatedSongs[0].MainArtist, FallbackValue={x:Null}, Mode=OneWay}"
```
huynhsontung, line 59: I'm surprised `x:Bind ViewModel.ContextAlbum.RelatedSongs[0].MainArtist` works tbh. We can't use Binding, so it has to be `x:Bind`. I also prefer the logic check in C# as you showed. So maybe:

```xml
CommandParameter="{x:Bind ViewModel.ContextAlbum.MainArtist, FallbackValue={x:Null}, Mode=OneWay}"
```
and 
```cs
public sealed partial class AlbumViewModel : ObservableRecipient
{
    ...
    public ArtistViewModel? MainArtist => RelatedSongs.Count > 0 ? RelatedSongs[0].MainArtist : null;
    ...
}
```

United600: Other than the count check they're basically the same, right? And I don't think that check really matters anyway. I even tried it with a file that had no tags, and it still navigated correctly to the "Unknown artist" page. But sure, this is similar to what I had before the item flyout bindings refactor.

United600: > I'm surprised x:Bind ViewModel.ContextAlbum.RelatedSongs[0].MainArtist works tbh.

Pretty nice right! Rebasing that older branch made me realize it was easier now, since the view model already exposes a context album, I could just pass it directly through the parameter without needing any extra properties.

huynhsontung (approved):

PR #994, opened 2026-07-23
refactor: resolve more Native AOT warnings

huynhsontung:
Mostly resolving nullable object references and event handler parameters not being nullable-aware.


huynhsontung:
@United600 This may trigger some merge conflicts. Just a heads-up.

United600:
No worries. I just find the `object?` part strange. Why is the compiler complaining about it? And does Visual Studio create the events with it now? Because it would be tedious to annotate them every time when creating them from XAML.

United600:
It'll be a great day when we can finally remove `#nullable enable`, and `partial` to some extent, from all the C# files 🎆.

huynhsontung:
> I just find the object? part strange. Why is the compiler complaining about it? And does Visual Studio create the events with it now?

I think it's because the CsWinRT generation is not fully nullable-aware. So the event signature is unclear to the compiler about whether the value can be null. The compiler picks a safer option and makes it nullable so the user can handle it inside the event handler. Again, hopefully CsWinRT 3 would resolve this.

PR #1002, opened 2026-07-27
feat: ChapterPickerControl handles empty chapter name

huynhsontung:
Before

image

After
image

### DIFF
--- a/Screenbox/Converters/ChapterTitleConverter.cs
+++ b/Screenbox/Converters/ChapterTitleConverter.cs
@@ -1,0 +1,42 @@
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+using Windows.Media.Core;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Data;
+
+namespace Screenbox.Converters;
+
+public sealed class ChapterTitleConverter : DependencyObject, IValueConverter
+{
+    public static readonly DependencyProperty ChaptersProperty = DependencyProperty.Register(
+    nameof(Chapters),
+    typeof(IList<ChapterCue>),
+    typeof(ChapterTitleConverter),
+    new PropertyMetadata(null));
+
+    public IList<ChapterCue>? Chapters
+    {
+        get => (IList<ChapterCue>?)GetValue(ChaptersProperty);
+        set => SetValue(ChaptersProperty, value);
+    }
+
+    public object Convert(object value, Type targetType, object parameter, string language)
+    {
+        return GetChapterTitle(value as ChapterCue);
+    }
+
+    public object ConvertBack(object value, Type targetType, object parameter, string language)
+    {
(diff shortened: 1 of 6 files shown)


United600:
Good catch. I was actually thinking about refactoring the "Show chapter name when available" setting to use that `Chapter {0}` resource, maybe something like "Show chapter title" or the opposite "Hide chapter title to avoid spoilers".

Another option is to simply hide the chapter name in the player and use an icon instead.

Screenbox/Converters/ChapterTitleConverter.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        return GetChapterTitle(value as ChapterCue);
+    }
+
+    public object ConvertBack(object value, Type targetType, object parameter, string language)
+    {
+        throw new NotImplementedException();
+    }
+
+    private string GetChapterTitle(ChapterCue? chapterCue)
```
United600: Maybe just make this method `public static`, skip the `IValueConverter` stuff, and drop the duplicate in `PlayerControls.xaml.cs`.

huynhsontung: That wouldn't work. The converter needs a reference to the `ChapterCue` list. If we use function binding, `ChapterPickerControl` would need to bind to the function in the `PlayerControls` instance (can't use static method) from inside a DataTemplate. I couldn't figure out how to x:Bind outside of the data template scope.

United600: 🤔 I see, we do need to pass the chapter list, but `ChaptersSource` isn't static. I think I can work around that... famous last words.

Screenbox/Converters/ChapterTitleConverter.cs
```diff
@@ -11,32 +11,43 @@ namespace Screenbox.Converters;
...
+            ? chapterCue.Title.TrimStart()
+            : Strings.Resources.ChapterName(chapters.IndexOf(chapterCue) + 1);
     }
 
-    public object ConvertBack(object value, Type targetType, object parameter, string language)
+    public object Convert(object value, Type targetType, object parameter, string language)
     {
-        throw new NotImplementedException();
+        return value is not ChapterCue chapter
+            ? DependencyProperty.UnsetValue
```
huynhsontung, line 45: Hmm why do we assume the receiver is a `DependencyProperty`? I don't even know what type the `UnsetValue` is. Why don't we just always return a string, like `string.Empty`?

United600: I just followed [this](https://learn.microsoft.com/en-us/windows/apps/develop/data-binding/data-binding-in-depth#value-converters):

"Note: If there is an error in the conversion, do not throw an exception. Instead, return DependencyProperty.UnsetValue, which will stop the data transfer."

PR #1005, opened 2026-07-28
refactor: notification view and view model

United600:
Continuation of #989, partially reverting more modifications made in #794 and completing the migration of view models to the Core project.

Not perfect, but acceptable for the moment.


huynhsontung (approved):
Thank you. `NotificationViewModel` can now reunite with `Screenbox.Core`. Balance has been restored.

PR #1007, opened 2026-07-31
feat: rework SelectionViewModel to operate on ranges

huynhsontung:
This PR overhauls `SelectionViewModel` to use a range-based architecture native to Windows `ListViewBase.SelectedRanges`. The new logic strictly tracks selections as index ranges rather than instantiating individual `SelectedItems`.

### `SelectionViewModel` Architecture
*   `SelectedRanges` is now exposed as a `ReadOnlyObservableCollection<ItemIndexRange>`. Consumers can rely directly on index ranges instead of individual items.
*   Refactored range compaction to occur entirely in-memory. The observable collection is only updated after ranges are merged and ordered.
*   `SelectedCount` is now a standard `[ObservableProperty]` updated deterministically during range mutations.
* Removed references to `DispatcherQueue` and `DispatcherQueueTimer` since all operations are synchronous.


Screenbox.Core/ViewModels/SelectionViewModel.cs
```diff
@@ -1,78 +1,65 @@
...
+    private int _selectedCount;
 
     /// <summary>
     /// Gets or sets a value that indicates whether all items are selected.
     /// </summary>
-    /// <value>
-    /// <see langword="true"/> if all items are selected; <see langword="false"/> if none
-    /// are selected; otherwise, <see langword="null"/> to indicate a mixed selection.
-    /// The default is <see langword="false"/>.
-    /// </value>
```
United600, line 40: Still relevant, so there's no need to remove it.

Screenbox.Core/ViewModels/SelectionViewModel.cs
```diff
@@ -1,78 +1,65 @@
...
     [ObservableProperty]
     private bool? _isAllSelected = false;
 
     /// <summary>
     /// Gets or sets a value that indicates whether selection mode is active.
     /// </summary>
-    /// <value>
-    /// <see langword="true"/> if selection mode is active; otherwise, <see langword="false"/>.
-    /// The default is <see langword="false"/>.
-    /// </value>
```
United600, line 50: Same as the previous observable property.

Screenbox.Core/ViewModels/SelectionViewModel.cs
```diff
@@ -85,53 +72,251 @@ public void SetItemsSource(IReadOnlyCollection<object>? source)
...
+        IsSelectionModeActive = newRanges.Count > 0;
+        _selectedRanges.SyncItems(newRanges);
+        SelectedCount = newRanges.Sum(r => (int)r.Length);
+        RefreshSelectionState();
+    }
+
+    /// <summary>
+    /// Compacts and normalizes the given ranges in memory without mutating <see cref="SelectedRanges"/>.
+    /// </summary>
+    private List<ItemIndexRange> CompactRangesInternal(IEnumerable<ItemIndexRange> ranges)
```
United600: nit: I'd move all the private methods to the bottom.

Screenbox/Behaviors/ListViewSelectionBehavior.cs
```diff
@@ -1,56 +1,35 @@
...
     /// <summary>
-    /// Identifies the <see cref="SelectedItems"/> dependency property.
+    /// Identifies the <see cref="Selection"/> dependency property.
     /// </summary>
-    public static readonly DependencyProperty SelectedItemsProperty = DependencyProperty.Register(
-        nameof(SelectedItems),
-        typeof(IEnumerable),
+    public static readonly DependencyProperty SelectionProperty = DependencyProperty.Register(
+        nameof(Selection),
+        typeof(SelectionViewModel),
```
United600, line 22: I'm not a fan of this, but it might be the only practical way forward, even though the behavior ideally shouldn't need to know anything about the view model.

huynhsontung: `SelectionViewModel` doesn't attach to a specific control or page, so at this point I don't think it's a view model, more like a helper class, like `ObservableCollection`. Maybe we should rename it to `ObservableSelection` and make it an `ObservableObject`?

United600: `ObservableSelection` seems fine, though I'm still not sure whether it should count as a view model.

huynhsontung: We can also have a contract for something like `ISelectionRange` that just exposes the `SelectedRanges` and `SetRanges()`. I don't know if that's an improvement.

PR #838, opened 2026-04-11
style: refactor empty state controls

United600:
Adds a new `ContentUnavailableView` control for displaying status or placeholder content in the UI, integrates it into the application, and adds supporting helpers for improved data binding.

https://github.com/user-attachments/assets/cc0da495-fcae-4026-a03a-a9b0667d5ad5

- Adds a `ContentUnavailableView` control that supports displaying a title, message, content, and action content, and is accessible for screen readers
- Adds a new "No playlists" `ContentUnavailableView` and updates the visual state instantaneously
- Adds `BooleanExtensions` and `CollectionExtensions` to provide extension methods for common XAML data binding usages
- Adds PlaylistOff glyph icon
- Refactors the welcome panel in `HomePage` to use the new `ContentUnavailableView` control
- Refactors the empty library in `MusicPage` to use the new `ContentUnavailableView` control
- Refactors the empty library in `VideosPage` to use the new `ContentUnavailableView` control
- Refactors the empty folder message in `FolderViewPage` to use the new `ContentUnavailableView` control
- Fixes the `HomePage` ContentStates visual state group to update instantaneously
- Fixes the issue of the welcome panel and empty libraries panels being obscured by the miniplayer

**Before**

https://github.com/user-attachments/assets/11ae03f4-ca94-45a4-bdc5-adb26e0cad65

**After**

https://github.com/user-attachments/assets/2a6bed42-0a17-48d5-a032-481b26614b2c

PR #1012, opened 2026-08-05
fix: improve playlists page gamepad navigation

United600:


United600:
The header split button is still broken (just like the one on the `HomePage`) and the root cause is the `IsNullOrEmptyStateTrigger` introduced in #838. That's why I didn't even bother attaching the `SplitButtonXY` behavior to it.
I already have a plan to fix that without removing the trigger, but I'm holding off until the separate controls project PR lands so I can address it in the new project structure.

PR #1013, opened 2026-08-05
feat: allow users to set a custom save location for frame captures

United600:
Closes #827. Inspired by #828, adds support for choosing a custom folder where frame captures are saved, instead of relying on a fixed default directory.

The design draws inspiration from both the **Gaming > Captures** "Captures location" and the equivalent setting used in the **Screen Recorder**. I added the `Open folder` button as a convenience feature for XBOX users.

Screenbox general settings

https://github.com/user-attachments/assets/3001298d-b96b-4e8e-b1be-30edd872f46f

### DIFF
--- a/Screenbox.Core/Services/FilesService.cs
+++ b/Screenbox.Core/Services/FilesService.cs
@@ -191,7 +191,12 @@ public sealed class FilesService : IFilesService
         return new MediaInfo(mediaType);
     }
 
-    public async Task<StorageFolder> GetFrameCaptureFolderAsync(string token)
+    public IAsyncOperation<StorageFolder> GetFrameCaptureFolderAsync(string token)
+    {
+        return GetFrameCaptureFolderFromTokenAsync(token).AsAsyncOperation();
+    }
+
+    private static async Task<StorageFolder> GetFrameCaptureFolderFromTokenAsync(string token)
     {
         if (!string.IsNullOrEmpty(token)
             && StorageApplicationPermissions.FutureAccessList.ContainsItem(token))
--- a/Screenbox.Core/Services/FilesService.cs
+++ b/Screenbox.Core/Services/FilesService.cs
@@ -191,6 +191,27 @@ public sealed class FilesService : IFilesService
         return new MediaInfo(mediaType);
     }
 
+    public async Task<StorageFolder> GetFrameCaptureFolderAsync(string token)
+    {
+        if (!string.IsNullOrEmpty(token)
+            && StorageApplicationPermissions.FutureAccessList.ContainsItem(token))
+        {
+            try
+            {
+                return await StorageApplicationPermissions.FutureAccessList.GetFolderAsync(token);
+            }
+            catch (Exception e) when (e is FileNotFoundException or UnauthorizedAccessException)
+            {
+                // Folder was moved, deleted, or access was revoked. Fall back to the default location.
+            }
+        }
+
+        StorageLibrary pictureLibrary = await StorageLibrary.GetLibraryAsync(KnownLibraryId.Pictures);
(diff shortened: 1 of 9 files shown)


Screenbox.Core/Services/FilesService.cs
```diff
@@ -191,6 +191,32 @@ public async Task<MediaInfo> GetMediaInfoAsync(StorageFile file)
         return new MediaInfo(mediaType);
     }
 
+    public IAsyncOperation<StorageFolder> GetFrameCaptureFolderAsync(string token)
```
United600, line 194: I based this approach on this Raymond Chen's [article](https://devblogs.microsoft.com/oldnewthing/?p=109955), and chose `IAsyncOperation<StorageFolder>` instead of `Task<Storage>` to stay aligned with the patterns used by `Windows.Storage.IStorageFolder` and WinRT.

Screenbox.Core/ViewModels/SettingsPageViewModel.cs
```diff
@@ -419,6 +425,38 @@ private async Task RemoveMusicFolderAsync(StorageFolder folder)
...
+        string oldToken = _settingsService.FrameCaptureFolderToken;
+        if (!string.IsNullOrEmpty(oldToken) && StorageApplicationPermissions.FutureAccessList.ContainsItem(oldToken))
+        {
+            StorageApplicationPermissions.FutureAccessList.Remove(oldToken);
+        }
+
+        string newToken = StorageApplicationPermissions.FutureAccessList.Add(folder);
+        _settingsService.FrameCaptureFolderToken = newToken;
+        FrameCaptureFolder = folder;
+    }
```
huynhsontung, line 444: `StorageApplicationPermissions.FutureAccessList` is used to send files to LibVLC and is cleared on every launch. The current logic in [PlayerService](https://github.com/huynhsontung/Screenbox/blob/20ff2f5ea0ee1651bdf96ab4224eb0aa428e9962/Screenbox.Core/Services/PlayerService.cs#L30-L39) should avoid clearing non-media files, but I haven't verified it. Can you confirm whether the save folder persists after many app re-launches?

United600: Oh, I hadn't considered that. From my testing, it has survived multiple restarts. How many restarts (and over what time span) are you thinking of?

huynhsontung: Should be fine. Small nit: When adding the folder to `FutureAccessList`, it's better to include a short string in the metadata to describe what it's for.

```cs
string newToken = StorageApplicationPermissions.FutureAccessList.Add(folder, "frame-capture");
```

Screenbox.Core/ViewModels/SettingsPageViewModel.cs
```diff
@@ -419,6 +425,38 @@ private async Task RemoveMusicFolderAsync(StorageFolder folder)
...
+        if (folder is null) return;
+
+        // Release the previous entry so the FutureAccessList (capped at 1000 items) doesn't accumulate stale folders.
+        string oldToken = _settingsService.FrameCaptureFolderToken;
+        if (!string.IsNullOrEmpty(oldToken) && StorageApplicationPermissions.FutureAccessList.ContainsItem(oldToken))
+        {
+            StorageApplicationPermissions.FutureAccessList.Remove(oldToken);
+        }
+
+        string newToken = StorageApplicationPermissions.FutureAccessList.Add(folder);
```
huynhsontung:
```suggestion
        string newToken = StorageApplicationPermissions.FutureAccessList.Add(folder, "frame-capture");
```

PR #1016, opened 2026-08-07
refactor: PlayerPageViewModel KeyDown handlers to void return pattern

United600:
It also updates several key‑handling methods in the view model to use a cleaner, more modern C# style, and they're now fully consistent across the board.

### DIFF
--- a/Screenbox/Controls/PlayerElement.xaml
+++ b/Screenbox/Controls/PlayerElement.xaml
@@ -36,7 +36,6 @@
             <Border Background="Transparent" />
         </UserControl>
 
-        <!--  Use -1 bottom margin to avoid white line artifact  -->
         <vlcSharp:VideoView
             x:Name="VlcVideoView"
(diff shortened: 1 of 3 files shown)


Screenbox/Controls/PlayerElement.xaml
```diff
@@ -39,7 +39,6 @@
         <!--  Use -1 bottom margin to avoid white line artifact  -->
         <vlcSharp:VideoView
             x:Name="VlcVideoView"
-            Margin="0,0,0,-1"
```
United600, line 42: I'm fairly sure b37b9c4 made this obsolete, including the bit in `ResizeKeyDown` where we compensated for the off‑by‑one visible‑bounds height:

```csharp
// Visible bounds always have 1 pixel less than actual window height?
var currentSize = new Size(view.VisibleBounds.Width, view.VisibleBounds.Height + 1);
```

PR #1024, opened 2026-08-09
style: add new SplitButtonEx templated control

United600:
Fixes an issue introduced in #838 that caused `SplitXYNavigationBehavior` and `SplitButtonExtensions` to stop working on pages using the `IsNullOrEmptyStateTrigger` trigger.

Integrates those adjustments applied to the WinUI2 `SplitButton` into a unified templated control and moves the custom styles and thickness converter to the controls project.


Screenbox.UI.Controls/XamlControlsResources.xaml.cs
huynhsontung, line 1: Why does `XamlControlsResources.xaml` have code behind? Isn't it just a `ResourceDictionary` with no binding?

United600: It's mainly there to keep the merge in `App.xaml` "clean". If this feels messy or not ideal, we can simplify things in one of two ways: use only `XamlControlsResources.xaml` (worse `App.xaml` merge), or use only `XamlControlsResources.cs` (like the approach used in [this](https://github.com/Team-FLUI/FluiFramework/blob/master/FluiFramework.UWP.Core/FluiFrameworkThemeResources.cs), which I pulled inspiration from in the first place).

Sorry, I was trying to be clever and get the best of both worlds.

United600: I just tested both approaches and although the latter gives a "cleaner" `App.xaml` merge, it ends up sacrificing IDE auto-completion.

huynhsontung: Oh I see the logic. I didn't fully understand it before.

huynhsontung (approved):

Screenbox.UI.Controls/XamlControlsResources.xaml.cs
United600: It's a tradeoff, if this merging doesn't feel "messy", we can just drop the `XamlControlsResources.xaml.cs` file.

```xml
<ResourceDictionary Source="ms-appx:///Screenbox.UI.Controls/XamlControlsResources.xaml" />
```

PR #1027, opened 2026-08-10
feat: use Windows SQLite3 lib instead of bundling a new one

huynhsontung:
Save like ~2MB. What can go wrong? :)


United600:
I like that, but I'm concerned about how outdated the built-in package is.

huynhsontung:
This is probably a bad idea. The `winsqlite3` only gets updated with Windows updates. Each Windows may have a different SQLite version. I don't know how compatible it is with our current database file. Plus, I thought it would save more. 2MB of storage is not even worth the risk.

PR #1030, opened 2026-08-10
refactor: replace TagLibSharp with WinRT

United600:


Screenbox.Core/ViewModels/MediaViewModel.cs
```diff
@@ -333,51 +335,19 @@ public async Task LoadThumbnailAsync()
...
-            if (cover == null) return null;
-            if (cover.Data.IsEmpty)
-            {
-                if (cover is not ILazy or ILazy { IsLoaded: true }) return null;
-                ((ILazy)cover).Load();
-            }
+            // Use SingleItem mode to retrieve embedded album art with original aspect ratio.
+            // https://learn.microsoft.com/windows/apps/develop/files/thumbnails
+            StorageItemThumbnail? source =
+                await file.GetThumbnailAsync(ThumbnailMode.SingleItem, requestedSize: 1280, ThumbnailOptions.UseCurrentScale);
```
huynhsontung, line 346: I don't think WinRT would parse the file and return the Thumbnail on the spot for us. I think it just fetches the cached thumbnail and returns whatever size is available. Were you able to extract high-quality covers from this?

Screenbox.Core/ViewModels/MediaViewModel.cs
```diff
@@ -282,9 +280,11 @@ public async Task LoadDetailsAsync(IFilesService filesService)
             AltCaption = string.Empty;
     }
 
-    public async Task LoadThumbnailAsync()
+    public async Task LoadThumbnailAsync(int decodePixelHeight = 256)
     {
-        if (Thumbnail != null) return;
+        if (Thumbnail is not null)
+            return;
```
huynhsontung, line 286: How do we know to reload the thumbnail if the current size doesn't match the requested size? Do we want to handle a customizable size at all?

huynhsontung:
I don't think we can get away with not using a tag parsing library for long. We will have to support tag editing at some point. We may not need it for album art extraction, but we will for audio tag editing.

Screenbox.Core/ViewModels/MediaViewModel.cs
United600: I think I was, if you check the link in that code block, it says that with a certain combination it will return it: `"If the file has album art, then the thumbnail is the album art with the original aspect ratio of the file."`.

I added a `Debug.WriteLine` to confirm, and it does print the original image size. I might try tossing in even larger covers to see what Debug spits out, I'll probably have to bump the `requestedSize` once I do.

Screenbox.Core/ViewModels/MediaViewModel.cs
United600: Oh, good question. That's probably why I end up with a "stale" decoding height on the first play in the `PlayerPage` album art. And just to clarify, decoding size doesn't affect how thumbnails are retrieved, it only changes the "level of detail" used when the image is actually decoded.

United600:
> I don't think we can get away with not using a tag parsing library for long. We will have to support tag editing at some point. We may not need it for album art extraction, but we will for audio tag editing.

Can't [this API](https://learn.microsoft.com/uwp/api/windows.storage.fileproperties.musicproperties.savepropertiesasync) write the same properties it lets us read? For sure, in the long run using a dedicated tag library would definitely give us more flexibility.

huynhsontung:
Overall, I like this approach. Knowing we can write music properties to files using WinRT sells it. If we can confirm the high quality cover (up to 1024x1024) can always be extracted then I have no concern remove TagLibSharp.

United600:
It outputs images larger than 1024, and most of the time the size matches the original album art. But for album arts that are 3000×3000, even raising the requested size from 1280 to 2560 doesn't produce a higher‑quality version (likely because that larger image isn't cached).
  
It's unclear whether the issue comes from the documentation or the implementation. Ideally, the `requestedSize` parameter shouldn't be necessary at all 🤔.

https://github.com/user-attachments/assets/d5158794-e322-4796-89bf-51a13f05fd86

huynhsontung:
So 1920x1920 looks like the upper limit. Do you think it's enough for most use cases?

United600:
> So 1920x1920 looks like the upper limit.

It can go up to 2560x2560.

> Do you think it's enough for most use cases?

I think it is, but I don't know if it's efficient compared to alternatives.

huynhsontung:
WinRT parses the image data then uses it as a BitmapSource. The bitmap data doesn't have to travel across the ABI boundary so I think it should be more efficient.

PR #1031, opened 2026-08-10
fix: group headers item screen reader name

United600:


Screenbox.Core/Helpers/ObservableAlbumGroup.cs
```diff
@@ -26,4 +26,9 @@ public ObservableAlbumGroup(IGrouping<string, AlbumViewModel> group) : base(grou
     {
         Key = group.Key;
     }
+
+    public override string ToString()
+    {
+        return Key;
```
United600, line 32: Too bad we can't do this due to the method being in the Core project, which prevents localization and we don't really have much of a choice 🥲.

```csharp
return $"{Key}; {Count} {(Count == 1 ? "album" : "albums")}";
```

huynhsontung: Use a converter lol

United600: AFAIK, we can't. This control is unusual in that it doesn't inherit any of the `AutomationProperties` from its content. I've been trying to fix it for so long without success 😩, but maybe there's some other workaround.

huynhsontung (approved):

PR #1032, opened 2026-08-11
chore: manually reference CsWinRT 2.3.1

huynhsontung:
Manually reference the CsWinRT NuGet to get more up-to-date features rather than waiting for an SDK update.

Ref: https://discord.com/channels/372137812037730304/1360013646050099280/1536106749935882290

### DIFF
--- a/Screenbox.Core/ViewModels/FolderViewPageViewModel.cs
+++ b/Screenbox.Core/ViewModels/FolderViewPageViewModel.cs
@@ -30,7 +30,7 @@ public partial class FolderViewPageViewModel : ObservableRecipient,
     // so we need to maintain a separate list of StorageFolder for navigation.
     public string[] Breadcrumbs { get; private set; } = [];
 
-    public IReadOnlyList<StorageFolder> BreadcrumbLocations { get; private set; } = [];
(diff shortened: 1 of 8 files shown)


Screenbox.Core/ViewModels/FolderViewPageViewModel.cs
```diff
@@ -30,7 +30,7 @@ public partial class FolderViewPageViewModel : ObservableRecipient,
     // so we need to maintain a separate list of StorageFolder for navigation.
     public string[] Breadcrumbs { get; private set; } = [];
 
-    public IReadOnlyList<StorageFolder> BreadcrumbLocations { get; private set; } = [];
+    public StorageFolder[] BreadcrumbLocations { get; private set; } = [];
```
United600, line 33: Do you think this fixes the `BreadcrumbBar` issues when using `x:Bind` with `StorageFolder` data type?

huynhsontung: Nope. Tested. Still broken.

United600: Was worth the try, I guess 😩

PR #1034, opened 2026-08-11
chrore: remove redundant nullable directives

United600:


huynhsontung:
Somewhat relevant, do we want to run a mass code clean up on the Core project now that we have NAOT? This would fix issues like file scope namespace and sorted usings.

United600:
Why limit it to just the Core project? Still, it's a great idea.

I don't know the reason for this, but we should also probably change it:

https://github.com/huynhsontung/Screenbox/blob/8315968076def67b9639ab9b9cd37427ae290545/Screenbox.Core/Common/IPropertiesDialog.cs#L4

PR #1040, opened 2026-08-15
chore: run code cleanup on the entire solution

huynhsontung:


United600 (approved):
I didn't expect so many files to still not be file-scoped.

PR #1049, opened 2026-08-16
refactor: player OSD messaging and status badges

United600:
Refactors player status overlays to a single OSD message model, standardizes status formatting, and adds badge behavior to keyboard/gamepad interactions. The previous implementation was too spread out, which made the OSD behavior harder to maintain and extend. This refactor keeps status text, icon, and command type in one place, reducing duplication and making future playback notifications easier to add consistently.

- Adds a `PlaybackCommandKind` enum to describe player actions such as play/pause, rewind/fast-forward, subtitle change, etc.
- Replaces previous status messages (`UpdateStatusMessage`, `UpdateVolumeStatusMessage`, `ShowPlayPauseBadgeMessage`) with a unified `PlayerOsdUpdateMessage`
- Removes obsolete status-message types and related redundant logic.
- Refactors the toggle subtitle keyboard accelerator logic, by cleanly separating the view layer from the view‑model implementation
- Refactors seek message helper methods to use a `StringBuilder` instead of string interpolation
- Refactors the message status update logic so that most of the formatting is now handled in the view layer
- Updates the player badge animation so it aligns more closely with the status message animation, reducing visual clash and creating a more unified interaction

https://github.com/user-attachments/assets/e9b22408-d8e7-40d1-9fd4-65cef7a7fec6


Screenbox.Core/Messages/PlayerOsdUpdateMessage.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// <summary>
+    /// Creates a <see cref="PlayerOsdUpdateMessage"/> instance with the <see cref="HasMessage"/>
+    /// property set to <see langword="true"/>.
+    /// </summary>
+    /// <returns>A new <see cref="PlayerOsdUpdateMessage"/> instance.</returns>
+    public PlayerOsdUpdateMessage WithMessage() =>
+        this with { HasMessage = true };
+
+    //public PlayerOsdUpdateMessage WithValue(object value) =>
+    //    this with { Value = value };
```
United600: This probably wasn't the best approach, likely should've been a class so the fluent style methods don't end up creating new instances each time.

Screenbox.Core/ViewModels/PlayerControlsViewModel.cs
```diff
@@ -36,6 +36,12 @@ public sealed partial class PlayerControlsViewModel : ObservableRecipient,
 
     public bool ShouldBeAdaptive => !IsCompact && SystemInformation.IsDesktop;
 
+    /// <summary>
+    /// Gets a value that indicates whether the current playback item has exactly one subtitle track.
+    /// </summary>
+    /// <value><see langword="true"/> if the current playback item has exactly one subtitle track; otherwise, <see langword="false"/>.</value>
+    public bool HasSingleSubtitleTrackCount => MediaPlayer?.PlaybackItem is { SubtitleTracks.Count: 1 };
```
United600, line 43: I'm not entirely sold on this workaround, the `C` keyboard accelerator probably should always open the composite picker flyout, instead of behaving like the hybrid it is right now.

Screenbox.Core/ViewModels/PlayerElementViewModel.cs
```diff
@@ -259,7 +259,11 @@ public void ProcessPointerWheelInput(int delta, bool isHorizontal)
         if (!isHorizontal)
         {
             int volume = Messenger.Send(new ChangeVolumeRequestMessage(delta > 0 ? 2 : -2, true));
-            Messenger.Send(new UpdateVolumeStatusMessage(volume));
+            Messenger.Send(
+                new PlayerOsdUpdateMessage(
+                    delta > 0 ? PlaybackCommandKind.VolumeUp : PlaybackCommandKind.VolumeDown,
```
United600: These delta checks don't really add anything for volume, we'd probably be better off just having `Volume` and `Mute` values instead.

Screenbox.Core/ViewModels/PlayerElementViewModel.cs
```diff
@@ -556,12 +564,23 @@ or MediaPlaybackState.Buffering
...
+
+        if (isDeltaPositive)
+            statusStrBuilder.Append('+');
+
+        statusStrBuilder.Append(Humanizer.ToDuration(delta))
+            .Append(')');
+
+        Messenger.Send(
+            new PlayerOsdUpdateMessage(
+                isDeltaPositive ? PlaybackCommandKind.FastForward : PlaybackCommandKind.Rewind,
```
United600, line 574: Another delta check that doesn't really add value, the only place where they matter is in the seek keyboard accelerator (which changes the player badge horizontal alignment and icon).

Screenbox/Pages/PlayerPage.xaml.cs
```diff
@@ -543,4 +543,34 @@ private void EscapeKeyboardAccelerator_OnInvoked(KeyboardAccelerator sender, Key
...
+
+    private string GetMessageText(PlaybackCommandKind kind, object? value)
+    {
+        return kind switch
+        {
+            //PlaybackCommandKind.Rewind or PlaybackCommandKind.FastForward when value is TimeSpan time => Humanizer.ToDuration(time),
+            PlaybackCommandKind.Rewind or PlaybackCommandKind.FastForward when value is string timeStr => timeStr,
+            PlaybackCommandKind.VolumeUp or PlaybackCommandKind.VolumeDown when value is int volume => Strings.Resources.VolumeChangeStatusMessage(volume),
+            PlaybackCommandKind.RateUp or PlaybackCommandKind.RateDown when value is double rate => $"{rate.ToString("0.##", CultureInfo.CurrentCulture)}\u00D7",
+            //PlaybackCommandKind.AspectRatio when value is string ratio => $"{Strings.Resources.AspectRatio}: {Converters.ResourceNameToResourceStringConverter.FromName(ratio)}",
```
United600: (I think) This makes it impossible to work correctly:

https://github.com/huynhsontung/Screenbox/blob/d37eeffc58b3e05e12866da3f9958507000f0a91/Screenbox/Converters/ResourceNameToResourceStringConverter.cs#L68

Maybe we just add a new method that leaves that behavior out entirely.

Screenbox/Pages/PlayerPage.xaml.cs
```diff
@@ -543,4 +543,34 @@ private void EscapeKeyboardAccelerator_OnInvoked(KeyboardAccelerator sender, Key
...
+    {
+        if (ViewModel.PlayerVisibility == PlayerVisibilityState.Minimal)
+            return HorizontalAlignment.Center;
+
+        return kind switch
+        {
+            PlaybackCommandKind.Rewind => HorizontalAlignment.Left,
+            PlaybackCommandKind.FastForward => HorizontalAlignment.Right,
+            _ => HorizontalAlignment.Center,
+        };
```
United600, line 581:
```suggestion
        return kind switch
        {
            _ when ViewModel.PlayerVisibility == PlayerVisibilityState.Minimal => HorizontalAlignment.Center,
            PlaybackCommandKind.Rewind => HorizontalAlignment.Left,
            PlaybackCommandKind.FastForward => HorizontalAlignment.Right,
            _ => HorizontalAlignment.Center,
        };
```

Which style should we prefer?

Screenbox.Core/ViewModels/PlayerControlsViewModel.cs
huynhsontung: I was trying to match YouTube's behaviour of toggling captions to the default/last language using `C`. Picking the subtitle language is not easy, so this is the quick solution I could think of. Not the best, for sure.

Screenbox.Core/Messages/PlayerOsdUpdateMessage.cs
huynhsontung: I'm a believer that a record should contain only primitives or small types. `object? Value` is too vague to be put in a record imo.

huynhsontung: We also lost the type information compared to the dedicated messages: `int` for the volume message, `string` for the status message, `bool` for the badge message. Shoving them into an object is a step backward.

Screenbox.Core/ViewModels/PlayerElementViewModel.cs
huynhsontung: I'm confused. What do you mean by this?

Screenbox.Core/ViewModels/PlayerControlsViewModel.cs
United600: It's fine, in the end it's just a single property.

Screenbox.Core/Messages/PlayerOsdUpdateMessage.cs
United600: I mean, the type information only matters for the view, that's where we determine what the message actually is. I agree that `object?` is less than ideal, but how else would you handle the multiple types the message can take?

Originally I used `double NumericValue` and `string TextValue`, but I eventually decided it was cleaner to consolidate everything into a single property. Do you prefer that approach? Personally, I find it much worse for binding to the view.

Screenbox.Core/ViewModels/PlayerElementViewModel.cs
United600: That is pointless, since the icon is determined by the value, and we don't perform any conditional logic depending on whether it's Up or Down.

It could just be:

```csharp
Messenger.Send(new PlayerOsdUpdateMessage(PlaybackCommandKind.Volume, Value: volume, HasMessage: true);
```

Screenbox.Core/Messages/PlayerOsdUpdateMessage.cs
huynhsontung: I don't have any alternatives. I just want to point out that moving to a single message means we lose typing information. It's a compromise. On another note, I still think this message should be a class, not a record.

Screenbox.Core/ViewModels/PlayerElementViewModel.cs
huynhsontung: Yeah, just passing the volume value makes sense to me. I don't mind having this logic at the sender or the receiver.

Screenbox/Pages/PlayerPage.xaml.cs
huynhsontung: Ooh the second style is smart. I like it

Screenbox.Core/Messages/PlayerOsdUpdateMessage.cs
United600: Yeah, I'm not a fan of the `object` either, that was the best I could come up with 🥲.

> On another note, I still think this message should be a class, not a record.

Do I keep using fluent methods, or should I drop them as well?

huynhsontung: I don't mind the fluent methods. I haven't used it anywhere before so I want to see how it looks like using it.

Screenbox.Core/ViewModels/PlayerPageViewModel.cs
```diff
@@ -201,43 +213,43 @@ public void Receive(PropertyChangedMessage<IMediaPlayer?> message)
...
 
-        _dispatcherQueue.TryEnqueue(() =>
+        // Don't show status message when player is not visible.
+        if (message.HasMessage && PlayerVisibility == PlayerVisibilityState.Visible)
         {
-            StatusMessage = message.Value;
-            if (message.Value is null || message.Duration == System.Threading.Timeout.InfiniteTimeSpan)
+            _dispatcherQueue.TryEnqueue(() =>
             {
-                _statusMessageTimer.Stop();
```
United600, line 214: Is this happening because the stop timer was removed? I'm struggling to get this fixed in an efficient way.

https://github.com/user-attachments/assets/b288eba5-6886-4a8a-be40-1991c9da3930

United600: 1e4605e should fix it.

Screenbox/Pages/PlayerPage.xaml.cs
huynhsontung: Why? The string value shouldn't have `#` or `?` prefixes.

huynhsontung:
If you seek quickly without waiting for the message to dismiss, the arrows on the screen don't reflect the keys I press. 

https://github.com/user-attachments/assets/20b528a7-5695-413f-8540-40ff40050764

United600:
> If you seek quickly without waiting for the message to dismiss, the arrows on the screen don't reflect the keys I press.

It's just an unfortunate side effect of how `ChangeTimeRequestMessage` works. As far as the OSD message is concerned, the behavior is "correct". The arrow is currently determined solely by the delta, and changing the seek direction doesn't reset it.

For example, if you fast forward twice and then rewind three times in quick succession (all within a single `ChangeTimeRequestMessage` result), you end up with deltas like **+5 +10 +5 0 -5** when the expected sequence would be **+5 +10 -5 -10 -15**.

I'll check how hard it is to achieve that behavior or find alternatives.

huynhsontung:
I see. I would not ship it as-is. It's very jarring when pressing an arrow key and it doesn't match the on-screen arrow. The delta should be irrelevant here. Once an arrow key is pressed, the position changes instantly. The OSD simply informs the user of what has already changed and provides visual feedback. The feedback should confirm the user's action. 

This behavior would be more logical if pressing the arrow keys only staged the position change, allowing the user to make incremental edits to the delta before 'committing' the change.

Screenbox/Pages/PlayerPage.xaml.cs
United600: Yeah, my hypothesis was a bit off, so it won't work on its own and as it is. I do have a fix in mind that should address it, if you think the message adds any value.

United600:
> I see. I would not ship it as-is. It's very jarring when pressing an arrow key and it doesn't match the on-screen arrow.

Agreed, I didn't look into it because I assumed it was out-of-scope for this PR.

> This behavior would be more logical if pressing the arrow keys only staged the position change, allowing the user to make incremental edits to the delta before 'committing' the change.

I'm not a fan of that behavior when seeking with keyboard/gamepad input, it should happen instantly.

huynhsontung:
> Agreed, I didn't look into it because I assumed it was out-of-scope for this PR.

Let's do another PR based on this one. I can help you stack them.

United600:
> > Agreed, I didn't look into it because I assumed it was out-of-scope for this PR.
> 
> Let's do another PR based on this one. I can help you stack them.

Actually the fix isn't too complicated, and if this pull request is mostly complete and won't require significant changes, I'd prefer to implement it here.

huynhsontung:
The arrow behaviour looks good now. My one minor request is to shorten the entrance animation duration. The exit animation duration is perfect. When an indicator provides visual feedback for my action, I want it to be snappier. The current duration gives me a slight pause after I press my key. There is no animation when quickly switching between arrows, which is good. Just the initial entrance animation is a tad too slow for my brain.

United600:
That's the consequence of syncing both message and badge animations. I'll mess around with it and see what I can do. If I shorten it too much, they’ll hang around forever, so maybe I just need to tweak the message duration 🤔.

huynhsontung:
The entrance animation for anything OSD should be short, for both message and badge animations.

huynhsontung:
The OSD entrance durations look good now. I noticed the message has a translucent border, but the badge does not. Is it intentional?

United600:
> The OSD entrance durations look good now. I noticed the message has a translucent border, but the badge does not. Is it intentional?

I wouldn't say it is intentional, it's just how it's always been, but it went unnoticed because they rarely shared their screen.
They should either have it or not, and I'd probably vote to remove it.

huynhsontung:
I agree. You are right. I never noticed that they look different from each other. Other than that, this PR is good.

huynhsontung (approved):

PR #1064, opened 2026-08-22
chore: bump ReswPlus version and refactor markup extension usage

United600:


Screenbox/Controls/PlayerControls.xaml
United600, line 92: This one, and the audio options flyout, will be fixed in #1063. The playback speed flyout can be addressed immediately afterward, while the aspect‑ratio flyout will require additional work. Then we can remove the `Extensions.cs` workaround.

PR #1066, opened 2026-08-23
chore: add license headers to UI library project

United600:
Add project-level `.editorconfig` and standard MIT license headers across `Screenbox UI` C# and XAML files.


huynhsontung:
I know a lot of repositories do this, but do we really need an MIT header for every file?

United600:
> I know a lot of repositories do this, but do we really need an MIT header for every file?

Not really. I just wanted to make it abundantly clear that it uses a different license than the one applied to the rest of the repository.

Should I adapt it to the Lively project instead?

PR #1069, opened 2026-08-25
fix: CultureInfo.CurrentCulture doesn't respect language override

huynhsontung:
NAOT no longer syncs the application language override `ApplicationLanguages.PrimaryLanguageOverride` with `CultureInfo.CurrentCulture` as .NET Native does. CurrentCulture now always return the system culture info. Thus, we need to set CurrentCulture at app launch to match the app language.

### DIFF
--- a/Screenbox/App.xaml.cs
+++ b/Screenbox/App.xaml.cs
@@ -47,7 +47,7 @@ sealed partial class App : Application
         _ = InitializeSentrySdk();
         UnhandledException += App_UnhandledException;
         CoreApplication.UnhandledErrorDetected += CoreApplication_UnhandledErrorDetected;
-        SyncCurrentCultureWithAppLanguage();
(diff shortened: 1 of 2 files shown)


Screenbox/App.xaml.cs
```diff
@@ -173,6 +174,21 @@ private static IServiceProvider ConfigureServices()
         return services.BuildServiceProvider();
     }
 
+    private static void SyncCurrentCultureWithAppLanguage()
```
United600: I would move this method into the `GlobalizationHelper.cs` file.

United600 (approved):

Screenbox/App.xaml.cs
huynhsontung: @copilot+mai-code-1.1-flash Please move the `SyncCurrentCultureWithAppLanguage` method into `GlobalizationHelper`.

PR #1072, opened 2026-08-25
perf(core): introduce RecentContext to optimize homepage recent items load time

huynhsontung:
## Context & Motivation

Previously, every time the user navigated to `HomePage.xaml` (e.g. switching back from Videos, Music, or Settings, or closing the Player overlay):
1. `HomePageViewModel` was transiently instantiated with a new, empty `Recent` collection.
2. The empty collection triggered the `Welcome` visual state in XAML (`IsNullOrEmptyStateTrigger`), momentarily flashing the full-page welcome empty-state banner and collapsing the grid header.
3. `OnLoaded()` queried the WinRT `MostRecentlyUsedList` and executed `GetFileAsync(token)` for every entry in parallel, incurring Windows Storage broker IPC calls across all items on every navigation.
4. Once files were resolved, items were populated into `Recent`, causing the `Welcome` state to dismiss and the UI to flash/reflow into the recent items grid.

## Key Changes & Impact

- **`RecentContext`**: Added a singleton observable state holder in `Screenbox.Core/Contexts` storing `Recent`, `PathToMruMappings`, `TokenToMediaMappings`, and `IsLoaded`.
- **DI Registration**: Registered `RecentContext` as a singleton in `ServiceHelpers.cs`.
- **`HomePageViewModel`**:
  - Bound `Recent` directly to `_recentContext.Recent` so the grid displays immediately on page load without visual reload or empty-state flashing.
  - Cached MRU token lookups to eliminate redundant `GetFileAsync` IPC broker calls for existing items.
  - Used in-place synchronization (`Recent.SyncItems(...)`) so zero collection change events or UI re-layouts occur when MRU entries are unchanged.
  - Avoided redundant thumbnail/metadata loading for already-loaded media.
- **`SettingsPageViewModel`**: Reused `ClearRecentHistory()` to reset `RecentContext` and `MostRecentlyUsedList` when clearing history or disabling `ShowRecent`.
- **Unit Tests**: Added `RecentContextTests.cs` covering initial state and property change notifications.

## Testing & Verification

- **Unit Tests**: Full test suite passed (30 / 30 tests) via `dotnet run --project Screenbox.Core.Tests/Screenbox.Core.Tests.csproj`.
- **Builds**:
  - Built `Screenbox.Core` via MSBuild (0 errors).
  - Built `Screenbox` via MSBuild (0 errors).
- **XAML Formatting**: Verified with `./scripts/lint-xaml.ps1`.

### DIFF
--- a/Screenbox.Core.Tests/Contexts/RecentContextTests.cs
+++ b/Screenbox.Core.Tests/Contexts/RecentContextTests.cs
@@ -1,0 +1,30 @@
+using Screenbox.Core.Contexts;
+
+namespace Screenbox.Core.Tests.Contexts;
+
+public class RecentContextTests
+{
+    [Test]
+    public async Task RecentContext_InitialState_ShouldBeEmptyAndNotLoaded()
+    {
+        var context = new RecentContext();
+
+        await Assert.That(context.Recent).IsEmpty();
+        await Assert.That(context.PathToMruMappings).IsEmpty();
+        await Assert.That(context.TokenToMediaMappings).IsEmpty();
+        await Assert.That(context.IsLoaded).IsFalse();
+    }
+
+    [Test]
+    public async Task RecentContext_IsLoaded_ShouldRaisePropertyChanged()
+    {
+        var context = new RecentContext();
+        string? changedProperty = null;
+        context.PropertyChanged += (s, e) => changedProperty = e.PropertyName;
+
+        context.IsLoaded = true;
+
+        await Assert.That(context.IsLoaded).IsTrue();
+        await Assert.That(changedProperty).IsEqualTo(nameof(RecentContext.IsLoaded));
+    }
+}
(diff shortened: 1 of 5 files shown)


United600:
While this is an improvement, I'm not sure we should compound on the current Recent implementation. The experience still feels a bit limited, especially when compared to the Media Player.

huynhsontung:
This PR aims only to improve the performance of the current implementation. Whether the current Recent is a good implementation is a different topic. How can we improve on the Home page experience?

United600:
> This PR aims only to improve the performance of the current implementation.

It's a great improvement, but I worry it adds complexity that will make future refactors more difficult.

> How can we improve on the Home page experience?

I'd refactor the `Recent` collection to accept an `IRecentItem` (or `IMediaItem`) instead of a `MediaViewModel`, so it would show albums, playlists, artists in addition to individual media items.

huynhsontung:
I like that idea. Since it will involve changes at every level, from the database to the view, we would pretty much have to rewrite the whole Recent logic, so I wouldn't worry too much about adding complexity when we get there.

United600:
> I like that idea. Since it will involve changes at every level, from the database to the view, we would pretty much have to rewrite the whole Recent logic, so I wouldn't worry too much about adding complexity when we get there.

You're right.

PR #1073, opened 2026-08-25
fix: refresh PlaylistDetailsPage button availability on collection changes

United600:
https://github.com/user-attachments/assets/ef7304ca-3ca5-4c07-aa4d-fe5e85625d96

### DIFF
--- a/Screenbox.Core/ViewModels/PlaylistDetailsPageViewModel.cs
+++ b/Screenbox.Core/ViewModels/PlaylistDetailsPageViewModel.cs
@@ -66,7 +66,11 @@ public sealed partial class PlaylistDetailsPageViewModel : ObservableRecipient
 
     private void SourceItems_CollectionChanged(object? sender, NotifyCollectionChangedEventArgs e)
     {
-        _commandRefreshTimer.Debounce(ShuffleAndPlayCommand.NotifyCanExecuteChanged, TimeSpan.FromMilliseconds(50));
+        _commandRefreshTimer.Debounce(() =>
+        {
+            PlayPlaylistCommand.NotifyCanExecuteChanged();
+            ShuffleAndPlayCommand.NotifyCanExecuteChanged();
+        }, TimeSpan.FromMilliseconds(50));
     }
 
     private static bool NotNull(MediaViewModel? item) => item != null;
@@ -79,6 +83,17 @@ public sealed partial class PlaylistDetailsPageViewModel : ObservableRecipient
         if (Source == null || item == null) return;
         var playlist = new Playlist(item, Source.Items);
         Messenger.Send(new SetQueueMessage(playlist, true));
+    }
+
+    [RelayCommand(CanExecute = nameof(NotEmpty))]
+    private void PlayPlaylist(PlaylistViewModel? playlist)
+    {
+        if (playlist?.Items is not { Count: > 0 })
+            return;
+
+        var firstItem = playlist.Items[0];
+        var queue = new Playlist(firstItem, playlist.Items);
+        Messenger.Send(new SetQueueMessage(queue, shouldPlay: true));
     }
 
     [RelayCommand(CanExecute = nameof(NotEmpty))]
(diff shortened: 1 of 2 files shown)


Screenbox.Core/ViewModels/PlaylistDetailsPageViewModel.cs
```diff
@@ -61,6 +85,17 @@ private void Play(MediaViewModel? item)
...
+    private void PlayPlaylist(PlaylistViewModel? playlist)
+    {
+        if (playlist?.Items is not { Count: > 0 })
+            return;
+
+        var firstItem = playlist.Items[0];
+        var queue = new Playlist(firstItem, playlist.Items);
+        Messenger.Send(new SetQueueMessage(queue, shouldPlay: true));
+    }
+
```
United600: I'm starting to think a new `ObservableProperty` might have been the right choice instead of a new `RelayCommand`.

huynhsontung: I'm not sure what you meant. How can `ObservableProperty` help in this situation?

United600: Instead of adding that command, we'd just expose a `FirstItem` observable property and bind that to the `CommandParameter`. Then we'd set `FirstItem` to `Source.Items.First()` whenever the collection changes or something along those lines. I think it would work, not sure it's actually better, though.

huynhsontung: Ah I get it now. True, you probably don't need a new `Command`. You reuse the `PlayCommand`. And since `CommandParameter` doesn't observe the value change, you could skip the property altogether and do something like:

```
CommandParameter="{linq:Enumerable.FirstOrDefault(ViewModel.Source)}"
```

I wonder if this works

huynhsontung: Nope. That did not work. Gotta go with the observable property route.

Screenbox.Core/ViewModels/PlaylistDetailsPageViewModel.cs
```diff
@@ -37,16 +44,39 @@ public PlaylistDetailsPageViewModel(
...
+        {
+            Source.Items.CollectionChanged += SourceItems_CollectionChanged;
+        }
+
+        FirstItem = Source?.Items.FirstOrDefault();
+    }
+
+    private void SourceItems_CollectionChanged(object? sender, NotifyCollectionChangedEventArgs e)
+    {
+        FirstItem = Source?.Items.FirstOrDefault();
```
United600, line 74: Why not place it directly inside the timer instead?

Screenbox.Core/ViewModels/PlaylistDetailsPageViewModel.cs
United600: Yeah, `x:Bind` and generics don't work well together. We could create our own that internally calls the LINQ method?

huynhsontung: I found out the command parameter value still needs to be observable. I would just create an ObservableProperty for binding rather than creating a LINQ wrapper that may not support observable collection notifications.

Screenbox.Core/ViewModels/PlaylistDetailsPageViewModel.cs
huynhsontung: I want to avoid timer blocks as much as possible since they use lambda and delay execution to another thread. You lose the current thread's context this way. I always prefer to run things synchronously or with awaited Tasks whenever possible. In this case, we have to use the timer to debounce the button flashing due to collection changes, so I'd put only the bare minimum code necessary to resolve it in the lambda.

United600: Understood. My concern is that with a playlist containing hundreds of items, any update would end up triggering that operation hundreds of times.

huynhsontung: Technically, it should trigger a notification only once, since the first item is the same across most operations. Subsequent sets just compare against the backing field and are a no-op. Observable properties only trigger notification when the new value doesn't equal to the existing one.

PR #1086, opened 2026-09-03
feat: add track and chapter navigation swipe gesture options

United600:
Screenbox swipe gesture settings

### DIFF
--- a/Screenbox.Core/Playback/VlcMediaPlayer.cs
+++ b/Screenbox.Core/Playback/VlcMediaPlayer.cs
@@ -522,6 +522,22 @@ public sealed class VlcMediaPlayer : IMediaPlayer
         }
     }
 
+    public void NextChapter()
+    {
+        if (PlaybackItem?.Media is null || PlaybackItem.Chapters.Count == 0)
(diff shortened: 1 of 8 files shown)


huynhsontung (approved):

Screenbox.Core/Playback/VlcMediaPlayer.cs
```diff
@@ -522,6 +522,22 @@ public void StepForwardOneFrame()
         }
     }
 
+    public void NextChapter()
+    {
+        if (PlaybackItem?.Media is null || PlaybackItem.Chapters.Count == 0)
+            return;
+
+        VlcPlayer.NextChapter();
```
United600, line 530: Neither this nor the `VlcPlayer.PreviousChapter` update the position when called while playback is paused. I attempted to replicate the approach used in `StepForwardOneFrame`, but the result is always off by one.

https://github.com/user-attachments/assets/67e6224d-e964-4e51-b370-dc6104ab338b

Finding the chapter start time by its index seems unnecessarily complex for this.

PR #1091, opened 2026-09-08
chore: NAOT build cleanup and minor CI update

United600:
- Ensures MSIX packages are produced only during CI by checking `ContinuousIntegrationBuild`, preventing unnecessary packaging and deployment failures during local builds
- Restores the publish profiles
- Disables [ImplicitUsings](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#implicitusings) to stop the .NET SDK from auto‑adding using directives
- Updates `main.yml` GitHub Actions dependencies to their latest versions, and several minor workflow improvements
- Updates the `update-manifest` script to support installing [unsigned sideload package](https://learn.microsoft.com/en-us/windows/msix/package/unsigned-package#create-an-unsigned-package)
- Adds a `Update-ProjectVersion` script to update the project version and copyright
- Adds a `Set-Secrets` script, and replaces the previous two-step action with a single one

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -72,16 +72,21 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-nuget-
 
-      - name: Fetch Gomplate
-        env:
-          GH_TOKEN: ${{ github.token }}
-        run: gh release download v5.0.0 --repo hairyhenderson/gomplate --pattern 'gomplate_*amd64.exe' --output .\gomplate.exe
-
       - name: Render secrets file
         env:
-          APPCENTER_API_KEY: ${{ github.ref_type == 'tag' && secrets.APPCENTER_API_KEY || '' }}
           SENTRY_DSN: ${{ github.ref_type == 'tag' && secrets.SENTRY_DSN || '' }}
-        run: Get-Content Screenbox/Secrets.cs.template | .\gomplate.exe | Set-Content Screenbox/Secrets.cs
+        run: |
+          $template = @"
+        namespace Screenbox;
+
+        internal static class Secrets
+        {
+            public const string SentryDsn = "{{SENTRY_DSN}}";
+        }
+        "@
+
+          $output = $template.Replace("{{SENTRY_DSN}}", $env:SENTRY_DSN)
+          Set-Content -Path Screenbox/Secrets.cs -Value $output
 
       - name: Get tag version
         id: get-tag-version
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -105,7 +105,7 @@ jobs:
           SENTRY_ORG: starpine
           SENTRY_PROJECT: screenbox
           SENTRY_URL: https://sentry.io/
-          SENTRY_RELEASE : screenbox@${{ steps.get-tag-version.outputs.version }}
+          SENTRY_RELEASE: screenbox@${{ steps.get-tag-version.outputs.version }}
           SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
         run: |
           if ('${{ github.ref_type }}' -eq 'tag') {
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -139,10 +139,8 @@ jobs:
         with:
           name: Screenbox-sideload
           path: |
-            C:/DeployOutput/**/*_Test/
-            !C:/DeployOutput/**/_Test/Dependencies/arm/**
-            !C:/DeployOutput/**/_Test/Dependencies/Win32/**
-            !C:/DeployOutput/**/*_Test/*.msix
+            C:\DeployOutput\*_Test\**
+            !C:\DeployOutput\*_Test\*.msix
           compression-level: 0
 
   release:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,157 +1,175 @@
-name:  Windows Release Builder
+name: Windows Release Builder
 
 on:
   push:
     branches:
-    - main
+      - main
     tags:
-    - v**
+      - v**
   pull_request:
 
 jobs:
   xaml-lint:
     runs-on: windows-latest
+    permissions:
+      contents: read
     steps:
-    - name: Checkout
-      uses: actions/checkout@v6
+      - name: Checkout
+        uses: actions/checkout@v7
 
-    - name: Setup dotnet
-      uses: actions/setup-dotnet@v5
-      with:
-        dotnet-version: '10.x'
+      - name: Install the .NET workload
+        uses: actions/setup-dotnet@v6
+        with:
+          dotnet-version: '10.x'
 
-    - name: Restore dotnet tools
-      run: dotnet tool restore
+      - name: Restore .NET tools
+        run: dotnet tool restore
 
-    - name: Lint XAML files
-      run: .\scripts\lint-xaml.ps1
+      - name: Run XAML style check
+        run: .\scripts\lint-xaml.ps1
 
   test:
     runs-on: windows-latest
+    permissions:
+      contents: read
     steps:
-    - name: Checkout
-      uses: actions/checkout@v6
+      - name: Checkout
+        uses: actions/checkout@v7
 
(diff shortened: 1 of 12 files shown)


.github/workflows/main.yml
```diff
@@ -105,12 +105,15 @@ jobs:
                         "-p:SentrySetCommits=true"
         }
         msbuild Screenbox\Screenbox.csproj `
+          -nologo `
+          -restore `
+          -v:minimal `
```
huynhsontung: Why do we change verbosity in CI? I think it's better for debugging if we keep it verbose right?

United600: Sure, makes sense.

.github/workflows/main.yml
```diff
@@ -1,154 +1,175 @@
...
+      - name: Fetch Gomplate
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: gh release download v5.0.0 --repo hairyhenderson/gomplate --pattern 'gomplate_*amd64.exe' --output .\gomplate.exe
+
+      - name: Render secrets file
+        env:
+          APPCENTER_API_KEY: ${{ github.ref_type == 'tag' && secrets.APPCENTER_API_KEY || '' }}
+          SENTRY_DSN: ${{ github.ref_type == 'tag' && secrets.SENTRY_DSN || '' }}
+        run: Get-Content Screenbox/Secrets.cs.template | .\gomplate.exe | Set-Content Screenbox/Secrets.cs
```
United600, line 84:
```suggestion
      - name: Render Secrets File
        env:
          APPCENTER_API_KEY: ${{ github.ref_type == 'tag' && secrets.APPCENTER_API_KEY || '' }}
          SENTRY_DSN: ${{ github.ref_type == 'tag' && secrets.SENTRY_DSN || '' }}
        run: |
          $content = Get-Content Screenbox/Secrets.cs.template -Raw
          $content = $content.Replace("{{APPCENTER_API_KEY}}", $env:APPCENTER_API_KEY)
          $content = $content.Replace("{{SENTRY_DSN}}", $env:SENTRY_DSN)
          Set-Content Screenbox/Secrets.cs $content
```

I was gonna change this into a PowerShell script, but I'm not comfortable with the risk of exposing secrets. There are probably safer ways to do this.

Also, can we remove the AppCenter stuff?

.github/workflows/main.yml
```diff
@@ -1,154 +1,175 @@
...
+      - name: Update package manifest
+        run: .\scripts\update-manifest.ps1 -Version "${{ steps.get-tag-version.outputs.version }}"
+
+      - name: Build app package
+        env:
+          PackageDir: C:\DeployOutput\
+          SENTRY_ORG: starpine
+          SENTRY_PROJECT: screenbox
+          SENTRY_URL: https://sentry.io/
+          SENTRY_RELEASE : screenbox@${{ steps.get-tag-version.outputs.version }}
```
United600, line 108:
```suggestion
          SENTRY_RELEASE: screenbox@${{ steps.get-tag-version.outputs.version }}
```

Is that trailing space intentional, or can I safely remove it without breaking anything?

Screenbox/Screenbox.csproj
```diff
@@ -27,13 +26,18 @@
     <AppxPackageSigningTimestampDigestAlgorithm>SHA256</AppxPackageSigningTimestampDigestAlgorithm>
     <GenerateTestArtifacts>False</GenerateTestArtifacts>
     <AppxAutoIncrementPackageRevision>False</AppxAutoIncrementPackageRevision>
-    <GenerateAppxPackageOnBuild>true</GenerateAppxPackageOnBuild>
+    <GenerateAppxPackageOnBuild Condition="'$(GITHUB_ACTIONS)' == 'true'">true</GenerateAppxPackageOnBuild>
```
United600, line 29: I think I've found a way to avoid passing the parameter into the library projects. Something like what they're doing [here](https://github.com/tig/winprint/blob/b37b7a6144dac60036e23fec4351eafeed6697f3/Directory.Build.props#L27) should work. Another question is whether we can avoid passing it to each project individually and make it a global setting instead.

United600:
I think we should set up self‑signing for the sideload package, following something along these [guidelines](https://learn.microsoft.com/en-us/windows/uwp/packaging/auto-build-package-uwp-apps#configure-package-signing) or [these](https://learn.microsoft.com/en-us/windows/apps/package-and-deploy/ci-for-winui3?pivots=winui3-packaged-csharp#step-3-set-up-your-workflow). This [one](https://learn.microsoft.com/en-us/windows/msix/desktop/azure-dev-ops) is probably less useful.

.github/workflows/main.yml
huynhsontung: This is fine. GitHub will automatically mask the value if any secrets are printed in the console. And yes, we can remove the AppCenter stuff.

.github/workflows/main.yml
huynhsontung: It's probably a copy-and-paste issue. Safe to remove.

Screenbox/Screenbox.csproj
huynhsontung: Instead of relying on `GITHUB_ACTIONS`, what about using the `ContinuousIntegrationBuild` property instead, so we don't reference any vendor-specific CI in the project file?

huynhsontung:
> I think we should set up self‑signing for the sideload package, following something along these [guidelines](https://learn.microsoft.com/en-us/windows/uwp/packaging/auto-build-package-uwp-apps#configure-package-signing) or [these](https://learn.microsoft.com/en-us/windows/apps/package-and-deploy/ci-for-winui3?pivots=winui3-packaged-csharp#step-3-set-up-your-workflow). This [one](https://learn.microsoft.com/en-us/windows/msix/desktop/azure-dev-ops) is probably less useful.

Uhh... we don't have a key file to sign the app with. That's why we have been relying on the Microsoft Store to sign it.

United600:
> Uhh... we don't have a key file to sign the app with. That's why we have been relying on the Microsoft Store to sign it.

I didn't mean it that way, just that after moving to .NET 10 the sideload artifact stopped being self‑signed with a temporary certificate. It made the package impossible to install.
With this PR the artifact can be installed unsigned, but only on Windows 11. So we should set up a signing workflow with a temporary certificate, like [this](https://learn.microsoft.com/en-us/windows/msix/package/sign-msix-package-guide) or [this](https://learn.microsoft.com/en-us/windows/msix/package/create-certificate-package-signing).

.github/workflows/main.yml
United600: I feel you. I made some in this PR too, and some of those are still hanging around. Not having automatic linting definitely doesn't help.

Screenbox/Screenbox.csproj
United600: Smart, I'll do that.

huynhsontung:
> > Uhh... we don't have a key file to sign the app with. That's why we have been relying on the Microsoft Store to sign it.
> 
> I didn't mean it that way, just that after moving to .NET 10 the sideload artifact stopped being self‑signed with a temporary certificate. It made the package impossible to install. With this PR the artifact can be installed unsigned, but only on Windows 11. So we should set up a signing workflow with a temporary certificate, like [this](https://learn.microsoft.com/en-us/windows/msix/package/sign-msix-package-guide) or [this](https://learn.microsoft.com/en-us/windows/msix/package/create-certificate-package-signing).

I see. Let's put a pin in that.

United600:
> > > Uhh... we don't have a key file to sign the app with. That's why we have been relying on the Microsoft Store to sign it.
> > 
> > 
> > I didn't mean it that way, just that after moving to .NET 10 the sideload artifact stopped being self‑signed with a temporary certificate. It made the package impossible to install. With this PR the artifact can be installed unsigned, but only on Windows 11. So we should set up a signing workflow with a temporary certificate, like [this](https://learn.microsoft.com/en-us/windows/msix/package/sign-msix-package-guide) or [this](https://learn.microsoft.com/en-us/windows/msix/package/create-certificate-package-signing).
> 
> I see. Let's put a pin in that.

Have a fix for it already, but it's still missing the release sideload package. Since it's tied to a larger restructuring of actions, I'll hold off it for now.

PR #1093, opened 2026-09-12
fix: add to playlist button in selection bar control

United600:
Fixes a regression introduced in #1007, where removing the `CommandParameter` from the usages of `SelectionBarControl` caused the `AddToPlaylistButton` to stop passing its context to the `AddToPlaylistFlyoutBehavior`.

We should probably rename the `SelectionBarControl.CommandParameter` property, as it no longer serves as a command parameter in the places where the control is used.


huynhsontung (approved):

huynhsontung:
> We should probably rename the `SelectionBarControl.CommandParameter` property, as it no longer serves as a command parameter in the places where the control is used.

Agree. Maybe a `SelectionBarControl.Selection` property?

United600:
> > We should probably rename the `SelectionBarControl.CommandParameter` property, as it no longer serves as a command parameter in the places where the control is used.
> 
> 
> 
> Agree. Maybe a `SelectionBarControl.Selection` property?

Seems fine, but we should stop binding the property to the other buttons if we go in that direction.

PR #1094, opened 2026-09-16
fix: lower OS min requirement from 10.0.19041 to 10.0.18362

huynhsontung:
OS min version was bumped from 10.0.18362 to 10.0.19041 during NAOT migration for a better chance of compatibility. Now that the migration is complete, we can lower the OS requirement back to the original.


Screenbox.Core.Tests/Screenbox.Core.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>net10.0-windows10.0.26100.0</TargetFramework>
-    <TargetPlatformMinVersion>10.0.19041.0</TargetPlatformMinVersion>
+    <SupportedOSPlatformVersion>10.0.18362.0</SupportedOSPlatformVersion>
```
United600, line 4: Nit: I would still use the `TargetPlatformMinVersion` property because it is still used in a blank UWP project, so it must be the correct way 🤷.

United600 (commented):
We should consider moving the `TargetPlatformMinVersion` and `TargetFramework` properties to the `Directory.Build.props` file. This way, we can centralize these settings and avoid manually configuring them for each project individually. For reference, you can see [this example](https://github.com/files-community/Files/blob/1a7c3dca812d53db7b1f7d00921a9dd45b37d706/Directory.Build.props#L6).

